PR #7, opened 2021-10-05
Config

aspacca:


handlers/aws/handler.py
```diff
@@ -4,36 +4,114 @@
...
+            logger.info("config file", extra={"bucket_name": bucket_name, "object_key": object_key})
+
+            config_storage: CommonStorage = StorageFactory.create(
+                storage_type="s3", bucket_name=bucket_name, object_key=object_key
+            )
+
+            config_yaml = config_storage.get_as_string()
+            logger.debug("config", extra={"yaml": config_yaml})
+
+            config: Config = parse_config(config_yaml)
```
jsoriano, line 50: Nit. For readability move the handling of each trigger type to their own methods. This may also help to avoid big try/except blocks.

handlers/aws/handler.py
```diff
@@ -4,36 +4,114 @@
...
+
+                    return
+
+            composite_shipper.flush()
+            logger.info("lambda processed all the events", extra={"sent_event": sent_event})
 
         except Exception as e:
+            logger.exception("exception raised", exc_info=e)
+            logger.error("lambda exited with an error", extra={"sent_event": sent_event})
             return str(e)
```
jsoriano, line 117: Nit. Instead of having so many try/except blocks intended to avoid the handler to raise exceptions, maybe we could have just a decorator that catches and logs all exceptions in the handler. We could still have smaller specific try blocks if needed for some specific case.

handlers/aws/sqs_trigger.py
```diff
@@ -4,35 +4,94 @@
...
             storage: CommonStorage = StorageFactory.create(
                 storage_type="s3", bucket_name=bucket_name, object_key=object_key
             )
 
-            for log_event, offset in storage.get():
+            span = elasticapm.capture_span(f"WAIT FOR OFFSET STARTING AT {last_ending_offset}")
+            span.__enter__()
+            for log_event, beginning_offset, ending_offset in storage.get_by_lines(
+                range_start=last_ending_offset, last_ending_offset=last_ending_offset
+            ):
```
jsoriano, line 73: Nit. Consider splitting long fors or if conditions with variables. Something like this here:
```suggestion
            events = storage.get_by_lines(
                range_start=last_ending_offset,
                last_ending_offset=last_ending_offset,
            )
            for log_event, beginning_offset, ending_offset in events:
```

handlers/aws/utils.py
```diff
@@ -2,30 +2,47 @@
...
-        "namespace": namespace,
-    }
+def from_s3_uri_to_bucket_name_and_object_key(s3_uri: str) -> tuple[str, str]:
+    if not s3_uri.startswith("s3://"):
+        raise ValueError(f"Invalid s3 uri provided: {s3_uri}")
+
+    s3_uri = s3_uri.strip("s3://")
 
-    event["event"] = {"dataset": dataset, "original": event["fields"]["message"]}
+    bucket_name_and_object_key = s3_uri.split("/", 1)
```
jsoriano, line 16: Nit. Use urllib to parse urls? https://docs.python.org/3/library/urllib.parse.html

PR #8, opened 2021-10-07
type check

francescayeye:


tests/requirements/lint-mypy.txt
```diff
@@ -0,0 +1,2 @@
+mypy
+types-PyYAML
```
jsoriano, line 2: We should pin the versions too to avoid issues in CI if new checks are added to the latest version of these linters.

jsoriano (approved):

tests/requirements/lint-mypy.txt
aspacca: on linting I would rather prefer to be always on latest version, to keep updating the source code to latest features/rules

aspacca: that's the whole point :)
I'd prefer to find changes in the linter during CI and fix ASAP on the limited piece of code that's affected, than pinning to a version and having to refactor much more code when and if the pinned version will be bumped

PR #9, opened 2021-10-15
Offset refactoring

aspacca:
- Added tests with coverage for the whole `storage` package
- Bugfixes in `S3Storage.get_by_lines` and `inflate` and `by_lines` decorators according to expectations in tests (handling newlines, offset of the lines in the files, range start when continuing the same file in another lambda execution)
- `inflate` and `by_lines` decorators refactored (after bugfixes applied) to wrapper functions instead of decorator classes since there's no need anymore to keep states in a class instance.
- CI and developers tools for running tests

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -51,43 +51,16 @@ pipeline {
     }
     stage('Test') {
       failFast false
-      matrix {
-        agent {label "${PLATFORM}"}
-        options { skipDefaultCheckout() }
-        axes {
-          axis {
-            name 'PLATFORM'
-            values 'ubuntu-20 && immutable' // , 'macosx && x86_64', 'windows-2019 && windows-immutable'
+      steps {
+        withGithubNotify(context: "Test") {
+          dir("${BASE_DIR}"){
+            sh(label: 'coverage', script: 'make coverage')
           }
         }
-        stages {
-          stage('build'){
-            steps {
-              withGithubNotify(context: "Build-${PLATFORM}") {
-                deleteDir()
-                unstash 'source'
-                dir("${BASE_DIR}"){
-                  echo 'TBD'
-                }
-              }
-            }
-          }
-          stage('Test') {
-            steps {
-              withGithubNotify(context: "Test-${PLATFORM}") {
-                deleteDir()
-                unstash 'source'
-                dir("${BASE_DIR}"){
-                  sh(label: 'coverage', script: 'make coverage')
-                }
-              }
-            }
-            post {
-              always {
-                junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/elastic-serverless-agent-junit.xml')
-              }
-            }
-          }
+      }
+      post {
+        always {
+          junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/elastic-serverless-agent-junit.xml')
         }
       }
     }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -61,6 +61,7 @@ pipeline {
       post {
         always {
           junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/elastic-serverless-agent-junit.xml')
+          cobertura(coberturaReportFile: "${BASE_DIR}/coverage.xml")
         }
       }
     }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -78,13 +78,13 @@ pipeline {
                 deleteDir()
                 unstash 'source'
                 dir("${BASE_DIR}"){
-                  echo 'TBD'
+                  sh(label: 'coverage', script: 'make coverage')
                 }
               }
             }
             post {
               always {
-                junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/junit-report.xml')
+                junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/elastic-serverless-agent-junit.xml')
               }
             }
           }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -54,7 +54,7 @@ pipeline {
       steps {
         withGithubNotify(context: "Test") {
           dir("${BASE_DIR}"){
-            sh(label: 'coverage', script: 'make coverage-ci')
+            sh(label: 'coverage', script: 'make coverage')
           }
         }
       }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -54,7 +54,7 @@ pipeline {
       steps {
         withGithubNotify(context: "Test") {
           dir("${BASE_DIR}"){
-            sh(label: 'coverage', script: 'make coverage')
+            sh(label: 'coverage', script: 'make coverage-ci')
           }
         }
       }
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,15 @@ help: ## Display this help text
 
 lint: black flake8 isort mypy ## Lint the project
 
+test:
+	# delete any __pycache__ folders to avoid hard-to-debug caching issues
+	find . -name __pycache__ -type d -exec rm -r {} +
+	py.test -v $(PYTEST_ARGS) $(PYTEST_JUNIT) tests
+
+coverage: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-branch
+coverage: export COVERAGE_FILE=.coverage
+coverage: test
+
 black:  ## Run black in the project
 	tests/scripts/docker/black.sh diff
 
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,9 @@ lint: black flake8 isort mypy ## Lint the project
 test:
 	PYTEST_ARGS="${PYTEST_ARGS}" tests/scripts/docker/run_tests.sh
 
-coverage: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-branch
+coverage: PYTEST_ARGS=--cov=. --cov-context=test --cov-config=.coveragerc --cov-branch
 coverage: export COVERAGE_FILE=.coverage
 coverage: test
-
-coverage-ci: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-branch
-coverage-ci: export COVERAGE_FILE=.coverage
-coverage-ci: test-ci
 
 black:  ## Run black in the project
 	tests/scripts/docker/black.sh diff
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,13 @@ coverage: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-br
 coverage: export COVERAGE_FILE=.coverage
 coverage: test
 
+test-ci:  ## Run test in the project
+	tests/scripts/docker/run_tests.sh
+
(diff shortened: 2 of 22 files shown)


v1v:
> @v1v is it possible to add codecov support?

IIRC, cobertura is the approach -> https://github.com/elastic/apm-agent-python/pull/735

Makefile
```diff
@@ -6,6 +6,17 @@ help: ## Display this help text
...
+test:
+	tests/scripts/docker/run_tests.sh
+
+coverage: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-branch
+coverage: export COVERAGE_FILE=.coverage
+coverage: test
+
+coverage-ci: PYTEST_ARGS=--cov --cov-context=test --cov-config=.coveragerc --cov-branch
+coverage-ci: export COVERAGE_FILE=.coverage
+coverage-ci: test-ci
```
v1v, line 18: `test-ci` is missing, therefore `coverage-ci` is duplicated, is that intentional?

Makefile
```diff
@@ -6,6 +6,17 @@ help: ## Display this help text
 
 lint: black flake8 isort mypy ## Lint the project
 
+test:
+	PYTEST_ARGS="${PYTEST_ARGS}" tests/scripts/docker/run_tests.sh
```
v1v, line 10: See 

```bash
+ py.test -v --cov --cov-context=test --cov-config=.coveragerc --cov-branch --junitxml=/app/tests/elastic-serverless-agent-junit.xml tests
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /app
plugins: cov-3.0.0
collecting ... collected 2 items

tests/storage/test_s3.py::TestS3Storage::test_get_as_string PASSED       [ 50%]
tests/storage/test_s3.py::TestS3Storage::test_get_by_lines PASSED        [100%]

------ generated xml file: /app/tests/elastic-serverless-agent-junit.xml -------

----------- coverage: platform linux, python 3.9.7-final-0 -----------
Name                        Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------------
tests/__init__.py               0      0      0      0   100%
tests/storage/__init__.py       0      0      0      0   100%
tests/storage/test_s3.py      101      0     10      0   100%
-------------------------------------------------------------
TOTAL                         101      0     10      0   100%

============================== 2 passed in 35.51s ==============================
+ coverage xml
$ ls -l coverage.xml 
-rw-r--r--  1 vmartinez  staff  4943 17 Oct 10:18 coverage.xml
```

tests/requirements/requirements.txt
```diff
@@ -1,6 +1,10 @@
...
-boto3>=1.7,<1.8
+boto3>=1.18,<1.19
 elasticsearch>=7.0,<8.0
-PyYAML>=5.3,<5.4
+PyYAML>=6.0,<6.1
 aws_lambda_typing>=2,<2.1
+mock>=4.0,<4.1
+pytest>=6.2,<6.3
+pytest-cov>=3.0,<3.1
+coverage==6.0.2
```
v1v, line 10: Required

tests/scripts/run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# delete any __pycache__ folders to avoid hard-to-debug caching issues
+find . -name __pycache__ -type d -exec rm -r {} +
+py.test -v ${PYTEST_ARGS} "${PYTEST_JUNIT}" tests
+coverage xml
```
v1v, line 8: Transform coverage to xml then Jenkins can parse it

.ci/Jenkinsfile
```diff
@@ -51,43 +51,17 @@ pipeline {
...
-              always {
-                junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/junit-report.xml')
-              }
-            }
-          }
+      }
+      post {
+        always {
+          junit(allowEmptyResults: true, keepLongStdio: true, testResults: '**/elastic-serverless-agent-junit.xml')
+          cobertura(coberturaReportFile: "${BASE_DIR}/coverage.xml")
```
v1v, line 64: Store the coverage output and show those details in the Jenkins UI.

In addition, those coverage reports are stored in ElasticSearch and some Kibana dashboards show them :)

- https://github.com/elastic/integrations/issues/755 as a reference

aspacca:
amazing @v1v , thanks!

storage/decorator.py
```diff
@@ -3,78 +3,88 @@
...
-        self._function: Callable[
-            [CommonStorageType, int, int, Any, str, int], Iterator[tuple[bytes, int, int]]
-        ] = function
-
-        self._offset: int = 0
-        self._last_ending_offset: int = 0
-
-    def __call__(self, instance: CommonStorageType, *args: Any) -> Iterator[tuple[bytes, int, int]]:
+def by_lines(
+    func: Callable[[CommonStorageType, int, Any, str, int], Iterator[tuple[bytes, int, int]]]
```
jsoriano, line 14: Would it be possible to define a type for this `Callable`? It looks error prone to use this in several places.

storage/decorator.py
```diff
@@ -3,78 +3,88 @@
...
-
-    def __call__(self, instance: CommonStorageType, *args: Any) -> Iterator[tuple[bytes, int, int]]:
-        for data, beginning_offset, ending_offset in self._function(instance, *args):
-            if args[3] == "application/x-gzip":
-                last_ending_offset: int = args[1]
+        _offset: int = 0
+        _range_start: int = 0
+
+        _newline: bytes = b""
+        _newline_length: int = 0
```
jsoriano, line 24: No need to use underscores to indicate privacy of variables local to a function.
```suggestion
        offset: int = 0
        range_start: int = 0

        newline: bytes = b""
        newline_length: int = 0
```

storage/decorator.py
```diff
@@ -3,78 +3,88 @@
...
+        iterator = func(storage, range_start, body, content_type, content_length)
+        while True:
+            try:
+                data, range_start, newline_length = next(iterator)
```
jsoriano, line 28: It looks a bit weird to need to run all the loop inside a try/except only to handle `StopIteration`.

Would it be possible to use this iterator in the loop? Something like this here:
```suggestion
        for data, range_start, newline_length in iterator:
```

And after the loop run the code in the current `except StopIteration`.

storage/decorator.py
```diff
@@ -3,78 +3,88 @@
...
+                if _offset == 0:
+                    _offset = _range_start
+
+                unfinished_line = unfinished_line + data
+                lines = unfinished_line.decode("UTF-8").splitlines()
```
jsoriano, line 33: How big can be this data? can it contain many lines? This is going to keep a copy in memory, splitted by lines.
An idea to avoid this may be to receive some kind of generator (aka a reader), and read the lines from there line by line. (But this may be some premature optimization if data is not so big).

tests/storage/test_factory.py
```diff
@@ -0,0 +1,34 @@
...
+
+import re
+from unittest import TestCase
+
+from storage import CommonStorage, S3Storage, StorageFactory
+
+
+class TestStorageFactory(TestCase):
+    def test_create(self) -> None:
+        with self.subTest("create s3 storage success"):
```
jsoriano, line 13: I didn't know this way of defining subtests in python, nice :slightly_smiling_face:

handlers/aws/sqs_trigger.py
```diff
@@ -28,7 +29,7 @@ def _handle_sqs_continuation(
     sqs_records = lambda_event["Records"][current_sqs_record:]
     body = json.loads(sqs_records[0]["body"])
     body["Records"] = body["Records"][current_s3_record:]
-    body["Records"][0]["last_ending_offset"] = last_ending_offset + 1
+    body["Records"][0]["last_ending_offset"] = last_ending_offset
```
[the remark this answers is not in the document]
aspacca: in the initial design we had both the `ending_offset` (the end of the last line processed) and `beginning_offset` (the char at where the line starts in the file). the latest is now calculated by the iterator consumer doing a diff between the ending offset, the line length and eventual newline length (1 or 2 for linux or windows).
all the values were affected by bugs indeed, and rather than missing events we were starting the continuation from a `last_ending_offset ` that was not really the char at were we left, but before.
regarding the `+1`, it is not needed since the char to start from while fetching the file again is zero based

storage/decorator.py
aspacca: oki for me

handlers/aws/sqs_trigger.py
```diff
@@ -92,10 +101,10 @@ def _handle_sqs_event(config: Config, event: dict[str, Any]) -> Iterator[tuple[d
                     span.__exit__(None, None, None)
                     span = None
 
-                es_event = _default_event.copy()
+                es_event = deepcopy(_default_event)
```
aspacca, line 104: `.copy()` is shallow, `_default_event` contains `fields` dict that would be referenced and updated on every iteration, ending with all the events collected by the shipper to have the partially the same content

share/logger.py
```diff
@@ -3,13 +3,16 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 
 import logging
+import os
 
 import ecs_logging
 from elasticapm.handlers.logging import LoggingFilter
 
+log_level = logging.getLevelName(os.getenv("LOG_LEVEL", logging.INFO))
```
aspacca, line 11: log level is `INFO` by default: it can be overridden by `LOG_LEVEL` env variable

storage/decorator.py
aspacca: consider as well, that while `text/plain` content we can stream, probably with `gzip` one we are bound to decompress everything in memory.

storage/decorator.py
```diff
@@ -3,75 +3,63 @@
...
 from share import shared_logger
 
-from .storage import CommonStorageType
+from .storage import CommonStorageType, GetByLinesCallable
 
 
-def by_lines(
-    func: Callable[[CommonStorageType, int, Any, str, int], Iterator[tuple[bytes, int, int]]]
-) -> Callable[[CommonStorageType, int, Any, str, int], Iterator[tuple[bytes, int, int]]]:
+def by_lines(func: GetByLinesCallable[CommonStorageType]) -> GetByLinesCallable[CommonStorageType]:
```
jsoriano, line 13: Nicer this way, yes :slightly_smiling_face: Is the `[CommonStorageType]` needed?

storage/decorator.py
```diff
@@ -3,75 +3,63 @@
...
-                            _newline = b"\n"
-                            _newline_length = len(_newline)
-
-                    if unfinished_line.endswith(_newline):
-                        unfinished_line = lines.pop().encode() + _newline
+        for data, range_start, _newline_length in iterator:
+            unfinished_line = unfinished_line + data
+            lines = unfinished_line.decode("UTF-8").splitlines()
+
+            if len(lines) > 0:
```
jsoriano, line 27: Nit. If there is nothing to do if there are no lines, maybe you can `continue` directly here. Saving in nesting levels uses to be a good idea in Python :slightly_smiling_face: 
```suggestion
            if len(lines) == 0:
                continue
```

jsoriano (approved):
Great, thanks!

storage/decorator.py
aspacca: because of `CommonStorage` being an abstract class: I have to use the related TypeVar/generic since the value passed to the decorators will be the concert implementation.
ie: passing the abstract class type is not enough to "catch" every concrete implementation

PR #11, opened 2021-10-20
Optimise storage memory usage

aspacca:
- Use `gzip.GzipFile` that supports `seek` method on a raw stream: this way we don't need to load the whole gzip content in memory
- CHUNK_SIZE lowered to 1K from 1M, the content is streamed anyway
- `by_lines` decorator still use buffer in memory since the biggest memory amount required will be up to the longest single line in the file plus a few bytes

### DIFF
--- a/storage/__init__.py
+++ b/storage/__init__.py
@@ -3,5 +3,5 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 
 from .factory import StorageFactory
-from .s3 import CHUNK_SIZE, S3Storage
-from .storage import CommonStorage
+from .s3 import S3Storage
+from .storage import CHUNK_SIZE, CommonStorage, StorageReader
--- a/tests/storage/test_s3.py
+++ b/tests/storage/test_s3.py
@@ -90,7 +90,7 @@ class TestS3Storage(TestCase):
     @mock.patch("storage.S3Storage._s3_client.head_object", new=MockContent.s3_client_head_object)
     @mock.patch("storage.S3Storage._s3_client.get_object", new=MockContent.s3_client_get_object)
     def test_get_by_lines(self) -> None:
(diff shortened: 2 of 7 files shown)


tests/storage/test_s3.py
```diff
@@ -32,20 +32,20 @@ def rewind() -> None:
 
     @staticmethod
     def init_content(newline: str) -> None:
-        MockContent.f_content = newline.join(
+        MockContent.f_content_plain = newline.join(
             [
                 "".join(random.choices(string.ascii_letters + string.digits, k=random.randint(0, 20)))
-                for _ in range(0, storage.CHUNK_SIZE)
+                for _ in range(0, 1024 ** 2)
```
endorama, line 38: Is possible to add a note on this magic number?

storage/__init__.py
```diff
@@ -3,5 +3,5 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 
 from .factory import StorageFactory
-from .s3 import CHUNK_SIZE, S3Storage
-from .storage import CommonStorage
+from .s3 import S3Storage
+from .storage import CHUNK_SIZE, CommonStorage, StorageReader
```
endorama, line 7: This change makes the `CHUNK_SIZE` smaller (from `1024 ** 2` in `.s3` to `1024` in `.storage`).

Is this change intended? I'm curious why `1024` is better than `1024 ** 2` in this context.

aspacca: `CHUNK_SIZE` is how much we read from the stream at every iteration. lowered from 1M to 1K since it will have less impact on memory and produce smaller iteration.
it doesn't seem to have any impact on speed (current `TestS3Storage.test_get_by_lines` takes around 10 sec with both values)

we will tweak the best value once we will have info about production data (avg/min/max filesize etc)

PR #14, opened 2021-10-27
support for cloud_id and api_key in elasticsearch client

aspacca:
New features:
- option to use CloudID instead of hosts/scheme in config in order to connect to Elasticsearch
- option to use Base64 encoded api key instead of username/password in config in order to authenticate to Elasticsearch


aspacca:
@endorama fixed

endorama (approved):

PR #24, opened 2021-10-29
Integration test

### LINKED ISSUE #16: Add integration tests to Elastic Server Forwarder

aspacca:
Closes #16

### DIFF
--- a/tests/handlers/aws/test_handler.py
+++ b/tests/handlers/aws/test_handler.py
@@ -17,7 +17,7 @@ from localstack.services.sqs.sqs_starter import check_sqs
 from localstack.utils import testutil
 from localstack.utils.aws import aws_stack
 
-from handlers.aws import lambda_handler
+from main_aws import handler
 
 
 class ContextMock:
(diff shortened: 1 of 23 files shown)


tests/handlers/aws/test_handler.py
```diff
@@ -127,104 +156,110 @@ def setUp(self) -> None:
...
+                        },
+                    ]
+                }
+
+                first_call = lambda_handler(event, ctx)  # type:ignore
+
+                assert first_call == "continuing"
+
+                while self._es_client.count(index="logs-redis.log-default")["count"] < 1:
+                    pass
```
endorama, line 216: How long does it usually take for this condition to be met? I was thinking to add a timeout here to prevent this test to hang up infinitely in case of issues and a sleep to go easier on the CPU.

tests/handlers/aws/test_handler.py
```diff
@@ -127,104 +156,110 @@ def setUp(self) -> None:
...
+                }
+                assert res["hits"]["hits"][0]["_source"]["fields"]["cloud"] == {"provider": "aws", "region": "eu-central-1"}
+
+                event = self._event_from_sqs_message()
+                second_call = lambda_handler(event, ctx)  # type:ignore
+
+                assert second_call == "continuing"
+
+                while self._es_client.count(index="logs-redis.log-default")["count"] < 2:
+                    pass
```
endorama, line 245: Same as before

tests/scripts/docker/run_tests.sh
```diff
@@ -6,7 +6,6 @@ docker_pip_cache="/tmp/cache/pip"
 
 cd tests
 
-docker network create run_tests || true
```
endorama: Is this network not needed any more? I've not been able to find where this network creation happens

tests/handlers/aws/test_handler.py
aspacca: there's a `timeout 5m` in the script running the test in docker:
https://github.com/elastic/elastic-serverless-forwarder/blob/f440f1b05021bc36b3780777673a18ab8464e5d1/tests/scripts/docker/run_tests.sh#L22

I'd rather force the timeout externally when running the test that explicitly declare the timeout in the test, since during the development and debugging I had to tweak it.

but open to consider different solution

tests/scripts/docker/run_tests.sh
aspacca: I had a first draft where I was bootstrapping `localstack` infra by code in the test, I needed for that to use the same network when running the test in docker (localstack spin up containers in the background).
this is not needed anymore since now I spin up the localstack container (I found out that localstack infra by code is not supported and deprecated) and expose the port on the host.
same for elasticsearch container.

aspacca:
@v1v what do you think about my solution of running docker in docker?

my goal was to be able to have no requirements in the integration test so that I could just execute it: whether on the IDE, with debugger session, or from CI etc

tests/handlers/aws/test_handler.py
aspacca: this test takes less than 1m running directly, a little more than 2mins in CI.

the main waiting time is starting the docker containers

aspacca: `self._es_client.count(index="logs-redis.log-default")["count"] < 1` in particular is going to wait approximately for `index.refresh_interval` value (https://www.elastic.co/guide/en/elasticsearch/reference/7.15/index-modules.html#index-refresh-interval-setting)

tests/scripts/run_tests.sh
```diff
@@ -4,7 +4,7 @@ set -ex
 
 # delete any __pycache__ folders to avoid hard-to-debug caching issues
 find . -name __pycache__ -type d -exec rm -r {} +
-py.test -v ${PYTEST_ARGS} "${PYTEST_JUNIT}" tests
+/home/user/.local/bin/py.test -v ${PYTEST_ARGS} "${PYTEST_JUNIT}" tests
```
v1v, line 7: I assume this particular script is not executed locally at all but always within the docker container context?

If so, what do you think to use `$HOME` rather than `/home/user`?

tests/scripts/run_tests.sh
```diff
@@ -4,7 +4,7 @@ set -ex
...
 # Transform coverage to xml so Jenkins can parse and report it
-coverage xml
-coverage html
+/home/user/.local/bin/coverage xml
+/home/user/.local/bin/coverage html
```
v1v, line 10: ditto

v1v:
> @v1v what do you think about my solution of running docker in docker?

I understand it's not ideal though. No strong opinion about this implementation. +1 in my case

tests/scripts/run_tests.sh
aspacca: yes, I will change to `$HOME` everywhere instead of `/home/user`

tests/handlers/aws/test_handler.py
aspacca: @endorama what do you think?

aspacca: I've force a `refresh` on the index, so I can now directly assert on the `count` api

endorama: Current solutions seems perfect! Nice work.

tests/scripts/run_tests.sh
endorama: As `$HOME/.local/bin` is in the `PATH`, is the absolute path needed here? Removing it would make this script usable from the outside too, given the tools are in place.

endorama (approved):
couple of minor comments but overall LGTM

tests/scripts/run_tests.sh
aspacca: `$HOME/.local/bin` is not in the `PATH`

aspacca: let's add it to the `PATH`

aspacca:
jenkins run the tests please

PR #26, opened 2021-11-04
find returns -1 when not found

aspacca:


endorama (approved):
If I understood correctly, if the searched string is at position `0` finds return `0`, thus not triggering the right condition. LGTM

aspacca:
> If I understood correctly, if the searched string is at position 0 finds returns 0, thus not triggering the right condition.

exactly!

it happens when the file contains an empty line

PR #29, opened 2021-11-08
Improve developer experience

aspacca:


aspacca:
rationale: separate docker run for CI and local run for developer that doesn't require build the docker image

tests/scripts/flake8.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+flake8 .
```
v1v, line 5: NIT: if this script is only calling `flake8 .` what do you think to remove it and add the call in the above consumers?

tests/scripts/mypy.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+mypy --install-types --non-interactive .
```
v1v, line 5: NIT: if this script is only calling `mypy` what do you think to remove it and add the call in the above consumers?

v1v (approved):
LGTM, some minor comments about removing a couple of files in favour or calling the command itself

tests/scripts/flake8.sh
aspacca: this way I can run the same script from docker or on the host.
If I move in the docker script then I cannot use the same script for both

tests/scripts/mypy.sh
aspacca: same as  above

tests/scripts/flake8.sh
aspacca: we could have in the consumer script the call to `flake8 .`, but still I'd need this script for running on the host and I have to maintain eventual changes in both place

PR #32, opened 2021-11-16
Esf review fixes

### LINKED ISSUE #15: Review of Elastic Serverless Forwarder Python code for best practices

aspacca:
fixes #15


aspacca:
missing README.md/CONTRIBUTING.md

shippers/factory.py
```diff
@@ -20,16 +20,16 @@ class ShipperFactory:
     @staticmethod
     def create_from_output(output_type: str, output: Output) -> CommonShipper:
         if output_type == "elasticsearch":
-            elasticsearch_output = ElasticSearchOutput(output.type, output.kwargs)
+            assert isinstance(output, ElasticSearchOutput)
```
masci, line 23: we should avoid `assert`ing at runtime, let's raise an exception instead

share/config.py
```diff
@@ -54,27 +53,13 @@ def __init__(self, output_type: str, kwargs: dict[str, Any]):
         self._dataset: str = ""
         self._namespace: str = ""
```
masci, line 54: let's remove the assignments above, the setters are invoked below and will take care of initializing the underlying fields

masci (changes requested):
Left a couple of comments, overall LGTM thanks for this PR!

shippers/factory.py
aspacca: the assertion is for type hinting so I can access "output.elasticserach_output_params" later on

do you suggest to add an if statement on `isinstance` check and raise an exception?
fine for me as long as I can achieve what I wrote above

aspacca:
@masci fixed

masci (approved):
LGTM, thanks!

PR #37, opened 2021-11-24
add license to bash script

aspacca:


kaiyan-sheng (approved):

endorama:
I'd like to suggest an alternative: https://github.com/elastic/go-licenser

Is the tool in use in beats and I think would be a better choice in the long run.

aspacca:
@endorama when I created the project https://github.com/elastic/go-licenser didn't support Elastic 2.0, that's the reason I wrote an ad hoc script

I see now it supports it, but It still has the problem that it puts the license before the shebang in bash :(

endorama:
Looking at `go-licenser` source it clearly does not support anything that is not Go code, so this solutions works for me. Meanwhile I [opened an issue](https://github.com/elastic/go-licenser/issues/43) to discuss possible expansion of file support.

PR #42, opened 2021-11-30
Secrets manager expansion

bogdan-copocean:
Added AWS SecretsManager to the config file

- It uses regex to scan the file, and searches for `"aws:arn:secretsmanager:REGION:ACCOUNT_ID:secret:SECRET_MANAGER_NAME:OPTIONAL_JSON_KEY"`
- Creates a list with all entries found
- Creates two dicts one with the raw_arn and one with the managers needed per region (in order to make the minimum number of requests to aws)
- Makes the requests to aws with the needed secrets managers with default version stage (AWSCURRENT)
- Parses the requests in order to match the needed keys with the fetched keys (the try/except block tries to load the json into a dict, and if it can't do that - because it's just a plain text string - it catches the error, and return the string)
- Populates the needed keys with the fetched keys
- Replaces every entry in the file with the fetched secret

### DIFF
--- a/share/config.py
+++ b/share/config.py
@@ -2,7 +2,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 
-from typing import Any, Optional
+from typing import Any, Callable, Optional
 
 import yaml
 
@@ -271,11 +271,14 @@ class Config:
         self._inputs[new_input.type][new_input.id] = new_input
 
 
-def parse_config(config_yaml: str) -> Config:
+def parse_config(config_yaml: str, expanders: list[Callable[[str], str]] = []) -> Config:
     """
     Config component factory
     Given a config yaml as string it return the Config instance as defined by the yaml
     """
+
+    for expander in expanders:
+        config_yaml = expander(config_yaml)
 
     yaml_config = yaml.safe_load(config_yaml)
     assert isinstance(yaml_config, dict)
--- a/share/secretsmanager.py
+++ b/share/secretsmanager.py
@@ -1,0 +1,167 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+import json
+import re
+from json.decoder import JSONDecodeError
+from typing import Any, Union
+
+import boto3
+from botocore.client import BaseClient as BotoBaseClient
+from botocore.exceptions import ClientError
+
+from .logger import logger as shared_logger
+
+
+def _get_aws_sm_client() -> BotoBaseClient:
+    """
+    Getter for secrets manager client
+    Extracted for mocking
+    """
+
+    return boto3.client("secretsmanager")
+
+
+def aws_sm_expander(config_yaml: str) -> str:
+    """
+    Secrets Manager expander for config file
+    It scans the file for the secrets manager arn pattern, checks for correct configuration,
+    retrieves the values from the secret manager and replaces them in the config file.
+    Exceptions will be risen for the following scenarios:
+        - Not respecting the arn pattern
+        - Input is for both plain text and json keys for the same secret manager name
+        - The fetched value is empty
+    """
+
+    config_secret_entry_values: dict[str, str] = {}
+    secret_key_values_cache: dict[str, dict[str, Any]] = {}
+    secret_consistency_len_check: dict[str, int] = {}
+
+    re_pattern = r"arn:aws:secretsmanager:(?:[^:]+):(?:[^:]+):secret:(?:[^\"']+)"
+    found_secrets_entries = re.findall(re_pattern, config_yaml)
+
+    for secret_arn in found_secrets_entries:
+        splitted_secret_arn = secret_arn.split(":")
+
+        if len(splitted_secret_arn) != 7 and len(splitted_secret_arn) != 8:
+            raise SyntaxError("Invalid arn format: {}".format(secret_arn))
+
+        if secret_arn not in config_secret_entry_values:
+            config_secret_entry_values[secret_arn] = ""
+
+        region = splitted_secret_arn[3]
+        secrets_manager_name = splitted_secret_arn[6]
+
+        if region == "":
+            raise ValueError("Must be provided region in arn: {}".format(secret_arn))
+
+        if secrets_manager_name == "":
+            raise ValueError("Must be provided secrets manager name in arn: {}".format(secret_arn))
+
+        if secrets_manager_name not in secret_consistency_len_check:
+            secret_consistency_len_check[secrets_manager_name] = len(splitted_secret_arn)
+        else:
+            if secret_consistency_len_check[secrets_manager_name] != len(splitted_secret_arn):
+                raise ValueError(
+                    "You cannot have both plain text and json key for the same secret: {}".format(secret_arn)
+                )
+
+        if region not in secret_key_values_cache:
+            secret_key_values_cache[region] = {}
+
+        if secrets_manager_name not in secret_key_values_cache[region]:
+            secret_key_values_cache[region][secrets_manager_name] = {}
+
+    for region in secret_key_values_cache:
+        for secrets_manager_name in secret_key_values_cache[region]:
+            str_secrets = get_secret_values(secrets_manager_name)
+            parsed_secrets = parse_secrets_str(str_secrets)
+
+            secret_key_values_cache[region][secrets_manager_name] = parsed_secrets
+
+    for config_secret_entry in config_secret_entry_values:
+        splitted_secret_arn = config_secret_entry.split(":")
+
+        region = splitted_secret_arn[3]
+        secrets_manager_name = splitted_secret_arn[6]
+
+        if len(splitted_secret_arn) == 8:
+            wanted_key = splitted_secret_arn[-1]
+            if wanted_key == "":
+                raise ValueError("Key must not be empty: {}".format(config_secret_entry))
+
+        if (
+            isinstance(secret_key_values_cache[region][secrets_manager_name], dict)
+            and wanted_key in secret_key_values_cache[region][secrets_manager_name]
+        ):
+            fetched_secret_entry_value = secret_key_values_cache[region][secrets_manager_name][wanted_key]
+            if fetched_secret_entry_value == "":
+                raise ValueError("Value for key: {} must not be empty".format(config_secret_entry))
+            config_secret_entry_values[config_secret_entry] = fetched_secret_entry_value
+        elif (
+            isinstance(secret_key_values_cache[region][secrets_manager_name], dict)
+            and wanted_key not in secret_key_values_cache
+        ):
+            raise KeyError("Key: {} not found in: {}".format(wanted_key, secrets_manager_name))
+        else:
+            config_secret_entry_values[config_secret_entry] = secret_key_values_cache[region][secrets_manager_name]
+
+        config_yaml = config_yaml.replace(config_secret_entry, config_secret_entry_values[config_secret_entry])
+
+    return config_yaml
+
+
+def get_secret_values(secret_name: str) -> str:
+    """
+    Calls the get_secret_value api from secrets manager, and returns the values.
+    If the secret is created in a binary format, it will be received as a byte string
+    on the "BinarySecret" key (boto3 does the base64 decoding internally).
+    Raises exceptions for ClientError errors.
+    """
+
+    secrets: str = ""
+    client = _get_aws_sm_client()
+
+    try:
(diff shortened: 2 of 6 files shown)


share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
         conf.add_input(current_input)
 
     return conf
+
+
+def aws_sm_expander(config_yaml: str) -> str:
+    needed_secret_arns = {}
+    needed_secret_keys = {}
+
+    re_pattern = r"arn:aws:secretsmanager:(?:[^:]*):(?:[^:]*):secret:(?:[^\n]*)"
```
aspacca, line 281: `[^\n]` is this for not matching a new line?
I would rather go for an ungreedy pattern/no multiline flag if python has support for those

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+    found_secrets = re.findall(re_pattern, config_yaml)
```
aspacca, line 282: nit: suffix `found_secrets` and `secret` etc below with `arn` to clarify what is the exact content

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+            needed_secret_keys[region][secrets_manager_name] = ""
+
+    for region in needed_secret_keys:
+        for secrets_manager_name in needed_secret_keys[region]:
+            str_secrets = get_secret_values(secrets_manager_name, region)
+            parsed_secrets = parse_secrets_str(str_secrets)
+
+            needed_secret_keys[region][secrets_manager_name] = parsed_secrets
+
+    for secret in needed_secret_arns:
```
aspacca, line 316: nit: `secret` has the same name here as before with a different content, let's consider a suffix to clarify what's the actual content here and avoid confusion

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+        elif e.response["Error"]["Code"] == "InvalidRequestException":
+            raise e
+        elif e.response["Error"]["Code"] == "ResourceNotFoundException":
+            raise KeyError("Cannot find secret: {}".format(secret_name)) from None
+    else:
+        if "SecretString" in get_secret_value_response:
+            secrets = get_secret_value_response["SecretString"]
+
+        else:
+            secrets = get_secret_value_response["SecretBinary"].decode("utf-8")
```
aspacca, line 364: this should be base64 decoded, shouldn't it?

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+
+
+def get_secret_values(secret_name: str, region_name: str) -> str:
+    session = boto3.session.Session()
+    client = session.client(service_name="secretsmanager", region_name=region_name)
+
+    try:
+        get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "DecryptionFailureException":
```
aspacca, line 349: let's add some debug logging in all these if branch :)

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+            raise ValueError("Value for key: {} must not be empty".format(secret))
+
+        config_yaml = config_yaml.replace(secret, needed_secret_arns[secret])
+
+    return config_yaml
+
+
+def get_secret_values(secret_name: str, region_name: str) -> str:
+    session = boto3.session.Session()
+    client = session.client(service_name="secretsmanager", region_name=region_name)
```
aspacca, line 344: this probably has to be moved to a function in order to be able to mock in tests

aspacca (commented):
just minor variables naming concerns

share/config.py
bogdan-copocean: should i add a separate function `get_aws_client(service_name: str, region_name: str) -> botocore.client.BaseClient` to handlers.aws.utils.py ? or do you want it somewhere above the get_secret_values?

aspacca: the implementation details are up to you

consider that in the unit tests and integration tests you will need to mock the client (see https://github.com/elastic/elastic-serverless-forwarder/blob/c85627cf4dbeb593ffd95868c647164634106487/tests/handlers/aws/test_handler.py#L260)

just for consistency instead of `session.Client` I would use https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
...
+                    needed_secret_arns[secret] = got_value
+
+        else:
+            needed_secret_arns[secret] = needed_secret_keys[region][secrets_manager_name]
+
+    for secret in needed_secret_arns:
+        if needed_secret_arns[secret] == "":
+            raise ValueError("Value for key: {} must not be empty".format(secret))
+
+        config_yaml = config_yaml.replace(secret, needed_secret_arns[secret])
```
aspacca, line 337: I wonder if we have both `aws:arn:secretsmanager:REGION:ACCOUNT_ID:secret:SECRET_MANAGER_NAME:OPTIONAL_JSON_KEY` and `aws:arn:secretsmanager:REGION:ACCOUNT_ID:secret:SECRET_MANAGER_NAME` and we replace the second one before the first one, then the second replace will fail.

This is clearly a misconfiguration (a secret is either plaintext or a json), but I think we should catch it upfront before running the input handler or sending data to the output, since failures at those stages might be harder to recover.

share/config.py
```diff
@@ -264,3 +272,107 @@ def parse_config(config_yaml: str) -> Config:
         conf.add_input(current_input)
 
     return conf
+
+
+def aws_sm_expander(config_yaml: str) -> str:
```
aspacca, line 277: nit: I would move this function and its helper to a separated module

share/config.py
bogdan-copocean: `shared_logger.debug("DecryptionFailureException: {}, secret: {}".format(e, secret_name))`  should we include the secret where it failed?

aspacca: yes, actually an `error` level would be better

prefer logging with payload as well, something like:
`shared_logger.error("get_secret_values", extra={"error_code": e.response["Error"]["Code"], "secret_name": secret_name})`

share/config.py
bogdan-copocean: totally agree, I will look for something that will match up to the double quotes and an optional last part (for the json key)

share/config.py
bogdan-copocean: it is correct that in the documentation it says it will be a base64 format, but when I tested it, it comes as a b"string". I will do more tests for this. Maybe I'm not creating the binary secret correctly.

share/config.py
bogdan-copocean: agreed, I will catch that too.

share/config.py
bogdan-copocean: I will look into it!

share/config.py
aspacca: maybe a base64 encoded as string type?
what's the content of the secret?

bogdan-copocean: accessing this: `get_secret_value_response["SecretBinary"]`, will output `b"this_is_my_secret"`. It's a bytes string. I don't need to decode from base64

aspacca: how was defined the secret?
can you try forcing a binary one with aws cli, thanks
```
aws secretsmanager put-secret-value --secret-id xxx --secret-binary fileb://asecret.file
```

bogdan-copocean: I defined the secret with create_secret api from boto3, but I also tried your way and it's still the same.. it returns the byte string

aspacca: https://github.com/boto/boto3/issues/2735

indeed boto3 does the decoding

share/config.py
bogdan-copocean: how does this look? `"arn:aws:secretsmanager:(?:[^:]*):(?:[^:]*):secret:(?:[^"|']*)(?::[^"|']*)?"`

after the `secret` it will match until it hits double quotes or single quotes, or it can have an optional `:JSON_KEY"` group, that will still match until hits single/double quotes.

bogdan-copocean: > 

i realized it's still greedy, still not what we want. working on it...

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+    It scans the file for the secrets manager arn pattern, checks for correct configuration,
+    retrieves the values from the secret manager and replaces them in the config file.
+    Exceptions will be risen for the following scenarios:
+        - Not respecting the arn pattern
+        - Input is for both plain text and json keys for the same secret manager name
+        - The fetched value is empty
+    """
+
+    needed_secret_arn: dict[str, str] = {}
+    needed_secret_keys: dict[str, dict[str, Any]] = {}
```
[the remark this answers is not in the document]
aspacca: `needed_secret_keys` this contains a key with the name of the secret extracted from the arn, does it?
and as value either the plain text secret, or the json secret, is it?

nit: I would rename to `secret_key_values_cache` or similar

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+    secret_consistency_len_check: dict[str, int] = {}
+
+    re_pattern = r"arn:aws:secretsmanager:(?:[^:]*?):(?:[^:]*?):secret:(?:.+?(?=\"|'))"
```
aspacca, line 37: I'm not quire sure every sub pattern needs to be ungreedy. just the last one in case
also in config the arn could not be surrounded by quotes (single or double)

what about:
`r"arn:aws:secretsmanager:(?:[^:]+):(?:[^:]+):secret:(?:[^"']+)"`
by default matching is not multiline, and we won't need to strip quotes after

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+    found_secrets_arn = re.findall(re_pattern, config_yaml)
```
aspacca, line 38: nit: `found_secrets_arn` it more like `found_secrets_entries`

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+            needed_secret_keys[region][secrets_manager_name] = {}
+
+    for region in needed_secret_keys:
+        for secrets_manager_name in needed_secret_keys[region]:
+            str_secrets = get_secret_values(secrets_manager_name)
+            parsed_secrets = parse_secrets_str(str_secrets)
+
+            needed_secret_keys[region][secrets_manager_name] = parsed_secrets
+
+    for empty_secret_arn in needed_secret_arn:
```
aspacca, line 80: `empty` as not set yet?

it is more like `config_secret_entry`

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+                raise ValueError("Invalid key requested from plain text secret: {}".format(empty_secret_arn))
+
+            for got_key, got_value in needed_secret_keys[region][secrets_manager_name].items():
+                if got_key == wanted_key:
+                    needed_secret_arn[empty_secret_arn] = got_value
+
+        else:
+            needed_secret_arn[empty_secret_arn] = needed_secret_keys[region][secrets_manager_name]
+
+    for parsed_secret_arn in needed_secret_arn:
```
aspacca, line 98: `parsed_secret_arn` is more like `config_secret_entry`

share/secretsmanager.py
```diff
@@ -0,0 +1,158 @@
...
+        if "SecretString" in get_secret_value_response:
+            secrets = get_secret_value_response["SecretString"]
+
+        else:
+            secrets = get_secret_value_response["SecretBinary"].decode("utf-8")
+
+    return secrets
+
+
+def parse_secrets_str(secrets: str) -> Any:
```
aspacca, line 144: `-> Union[str, dict[str, Any]]`

share/secretsmanager.py
```diff
@@ -0,0 +1,167 @@
...
+            if fetched_secret_entry_value == "":
+                raise ValueError("Value for key: {} must not be empty".format(config_secret_entry))
+            config_secret_entry_values[config_secret_entry] = fetched_secret_entry_value
+        elif (
+            isinstance(secret_key_values_cache[region][secrets_manager_name], dict)
+            and wanted_key not in secret_key_values_cache
+        ):
+            raise KeyError("Key: {} not found in: {}".format(wanted_key, secrets_manager_name))
+        else:
+            config_secret_entry_values[config_secret_entry] = secret_key_values_cache[region][secrets_manager_name]
```
aspacca, line 108: here `secret_key_values_cache[region][secrets_manager_name]` could be empty like for https://github.com/elastic/elastic-serverless-forwarder/pull/42/files#diff-db1d9ebcb636e71a1e8c74b68c0eb2aff1e7ca7e304450f805cbfcd7011c2ed2R99

we should raise a `ValueError` as well

aspacca (approved):

PR #44, opened 2021-12-07
Aws sm tests

bogdan-copocean:
1. Added region_name parameter to the secrets manager client, in order to access secrets from different regions.
2. Added unit tests for secrets manager

### DIFF
--- a/tests/handlers/aws/test_handler.py
+++ b/tests/handlers/aws/test_handler.py
@@ -15,11 +15,13 @@ import mock
 import pytest
 from elasticsearch import Elasticsearch
 from localstack.services.s3.s3_starter import check_s3
+from localstack.services.secretsmanager.secretsmanager_starter import check_secretsmanager
 from localstack.services.sqs.sqs_starter import check_sqs
 from localstack.utils import testutil
 from localstack.utils.aws import aws_stack
 
 from main_aws import handler
+from share.secretsmanager import aws_sm_expander
 
 
 class ContextMock:
@@ -122,6 +124,12 @@ class TestLambdaHandlerSuccess(TestCase):
         message["eventSourceARN"] = self._continuing_queue_info["QueueArn"]
         return dict(Records=[message])
 
+    def _create_secrets(self, secret_name: str, secret_data: dict[str, str]) -> None:
+        client = aws_stack.connect_to_service(
+            "secretsmanager", region_name="eu-central-1", endpoint_url=f"http://localhost:{self._LOCALSTACK_HOST_PORT}"
+        )
+        client.create_secret(Name=secret_name, SecretString=json.dumps(secret_data))
+
     @staticmethod
     def _upload_content_to_bucket(
         content: Union[bytes, str], content_type: str, bucket_name: str, key_name: str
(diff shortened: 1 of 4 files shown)


tests/share/test_secretsmanager.py
```diff
@@ -0,0 +1,239 @@
...
+                secret_data = secrets["data"]
+
+        return secret_data
+
+
+class TestAWSSecretsManager(TestCase):
+    @mock.patch(
+        "share.secretsmanager.get_secret_values",
+        new=MockContent.get_secret_values,
+    )
```
aspacca, line 90: I would rather mock `_get_aws_sm_client()` at https://github.com/elastic/elastic-serverless-forwarder/blob/8f80f6512bde1c2dcd5a7ba1d5bc515480baa3a7/share/secretsmanager.py#L17

or `client. get_secret_value` at https://github.com/elastic/elastic-serverless-forwarder/blob/8f80f6512bde1c2dcd5a7ba1d5bc515480baa3a7/share/secretsmanager.py#L129 (see https://github.com/elastic/elastic-serverless-forwarder/blob/main/tests/storage/test_s3.py#L87)

tests/share/test_secretsmanager.py
```diff
@@ -0,0 +1,239 @@
...
+                            password: "arn:aws:secretsmanager:eu-central-1:123-456-789:secret:es_secrets:password"
+                            dataset: "dataset"
+                            namespace: "namespace"
+            """
+            mock_fetched_data = aws_sm_expander(config_yaml)
+
+            parsed_config_yaml = """
+                inputs:
+                - type: sqs
+                    id: "aws:arn:secretsmanager:eu-central-1:123-456-789:secret:plain_secret"
```
aspacca, line 135: `aws:arn` instead of `arn:aws` is a little hard to spot out immediately :)

please add a comment about why the pattern does not match
like `#BEWARE aws:arn != arn:aws` :)

aspacca: is the pattern case sensitive?

bogdan-copocean: yes, it is. It will not match the pattern if you enter `Arn:aws...`

tests/handlers/aws/test_handler.py
```diff
@@ -258,103 +300,127 @@ def test_lambda_handler(self) -> None:
...
+                              - type: "elasticsearch"
+                                args:
+                                  elasticsearch_url: "http://127.0.0.1:{self._ES_HOST_PORT}"
+                                  username: "{self._ELASTIC_USER}"
+                                  password: "{self._ELASTIC_PASSWORD}"
+                                  dataset: "redis.log"
+                                  namespace: "default"
+                                """
+
+                        assert aws_sm_expander(self._config_yaml) == parsed_config_yml
```
aspacca, line 327: we can skip this in the integration test: we already have unit coverage for this.
the integration test should fail if the config is not parsed properly

tests/handlers/aws/test_handler.py
```diff
@@ -177,26 +189,53 @@ def setUp(self) -> None:
...
+        self._create_secrets("es_secrets", {"username": self._ELASTIC_USER, "password": self._ELASTIC_PASSWORD})
+
         self._elastic_container = docker_client.containers.run(
             "docker.elastic.co/elasticsearch/elasticsearch:7.15.1",
             detach=True,
             environment=[
                 "ES_JAVA_OPTS=-Xms1g -Xmx1g",
-                "ELASTIC_PASSWORD=password",
+                f"ELASTIC_PASSWORD={self._ELASTIC_PASSWORD}",
+                "xpack.security.enabled=true",
```
aspacca, line 218: 👍

tests/share/test_secretsmanager.py
```diff
@@ -0,0 +1,252 @@
...
+from unittest import TestCase
+
+import mock
+import pytest
+from botocore.exceptions import ClientError
+
+from share.secretsmanager import aws_sm_expander
+
+
+@pytest.mark.unit
```
aspacca, line 19: this should decorate `class TestAWSSecretsManager(TestCase):`, shouldn't it?

Issue #23, opened 2021-10-28
ESF: Custom tag support


aspacca:
1. config structure should accept a `tags` param as string list for output type `elasticsearch` (https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md#s3_config_file)
```
inputs:
  - type: "sqs"
    id: "arn:aws:sqs:%REGION%:%ACCOUNT%:%QUEUENAME%"
    outputs:
      - type: "elasticsearch"
        args:
          # either elasticsearch_url or cloud_id, elasticsearch_url takes precedence
          elasticsearch_url: "http(s)://domain.tld:port"
          cloud_id: "cloud_id:bG9jYWxob3N0OjkyMDAkMA=="
          # either api_key or username/password, apy_key takes precedence
          api_key: "YXBpX2tleV9pZDphcGlfa2V5X3NlY3JldAo="
          username: "username"
          password: "password"
          dataset: "generic"
          namespace: "default"
          tags:
            - tag1
            - tag2
            - tag3
```

2. `share.config.ElasticSearchOutput` should accept an extra `__init__` params for `tags` values coming from config
3. `shippers.es.ElasticsearchShipper` should accept an extra `__init__` params for `tags` values coming from config
4. `shippers.factory.ShipperFactory.create_from_output` should be adapted to pass `tags` for elasticsearch output type
5. `shippers.es.ElasticsearchShipper._enrich_event` should merge `tags` key for `event_payload` with the tags coming from input
6. Add unit test coverage for happy, sad and bad paths

7. Change the integration test `tests.handlers.aws.test_handler.TestLambdaHandlerSuccess.test_lambda_handler` to have at least one tags in the config file

8. Document the feature in `docs/README_AWS.md`

feel free to split further the ticket in multiple ones for every steps

PR #45, opened 2021-12-13
support for tags added

bogdan-copocean:
Tags support added for outputs (#23).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ tests/elastic-serverless-forwarder-junit.xml
 *.code-workspace
 .pytest_cache/
 .python-version
+htmlcov/
--- a/share/config.py
+++ b/share/config.py
@@ -83,8 +83,7 @@ class ElasticSearchOutput(Output):
             shared_logger.warning("no namespace set in config: using `default`")
             self.namespace = "default"
 
-        if not self.tags or len(self.tags) == 0:
-            shared_logger.warning("no tags set in config")
+        shared_logger.debug("tags: ", extra={"tags": self.tags})
 
     @property
     def elasticsearch_url(self) -> str:
@@ -172,7 +171,11 @@ class ElasticSearchOutput(Output):
         if not isinstance(values, list):
             raise ValueError("Tags must be of type list")
 
-        self._tags = [str(value) for value in values]
+        for value in values:
+            if not isinstance(value, str):
+                raise ValueError(f"All tags must be of type str: {value}")
+
+        self._tags = [value for value in values]
 
 
 class Input:
--- a/share/config.py
+++ b/share/config.py
@@ -44,6 +44,7 @@ class ElasticSearchOutput(Output):
         api_key: str = "",
         dataset: str = "",
         namespace: str = "",
+        tags: list[str] = [],
     ):
 
         super().__init__(output_type="elasticsearch")
@@ -54,6 +55,7 @@ class ElasticSearchOutput(Output):
         self.api_key = api_key
         self.dataset = dataset
         self.namespace = namespace
+        self.tags = tags
 
         if not self.cloud_id and not self.elasticsearch_url:
             raise ValueError("Elasticsearch Output elasticsearch_url or cloud_id must be set")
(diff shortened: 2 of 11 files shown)


share/config.py
```diff
@@ -81,6 +83,9 @@ def __init__(
             shared_logger.warning("no namespace set in config: using `default`")
             self.namespace = "default"
 
+        if not self.tags or len(self.tags) == 0:
```
aspacca, line 86: `tags` is not mandatory

share/config.py
```diff
@@ -81,6 +83,9 @@ def __init__(
...
+            shared_logger.warning("no tags set in config")
```
aspacca, line 87: no need for a `warning`, `debug` is fine.
while we are at it we can log the `tags`, regardless they are empty or not

tests/share/test_config.py
```diff
@@ -727,3 +843,4 @@ def test_parse_config(self) -> None:
             assert elasticsearch.api_key == "api_key"
             assert elasticsearch.dataset == "dataset"
             assert elasticsearch.namespace == "namespace"
+            assert elasticsearch.tags == ["2021", "{'key1': 'value1'}", "['elem1', 'elem2']"]
```
aspacca, line 846: good test case. but actually we should reject it :)
`tags` must be `list[str]`

bogdan-copocean: should we reject every tag? let's say the user inputs `"tag1", 2, "tag3"`. We're dropping all tags or only the `2`?

shippers/es.py
```diff
@@ -94,7 +96,7 @@ def _enrich_event(self, event_payload: dict[str, Any]) -> None:
 
         event_payload["event"] = {"dataset": self._dataset, "original": event_payload["fields"]["message"]}
 
-        event_payload["tags"] = ["preserve_original_event", "forwarded", self._dataset.replace(".", "-")]
+        event_payload["tags"] = self._tags
```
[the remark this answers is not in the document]
bogdan-copocean: merging like this: `["preserve_original_event", "forwarded", self._dataset.replace(".", "-")] + self._tags` ?

share/config.py
bogdan-copocean: removed the if statement and added: `shared_logger.debug("tags: ", extra={"tags": self.tags})` or `shared_logger.debug(f"tags: {self.tags}")`

tests/share/test_config.py
aspacca: I would reject everything: it's a validation error.

Btw: we parse the config in the handler and use `Output` for the shipper factory, we should add a bad path test on `tests.handlers.aws.test_handler.TestLambdaHandlerFailure.test_lambda_handler_failure` for tags failing validation.

we should have done for the secret manager as well, I will open a ticket for it :)

share/config.py
```diff
@@ -192,6 +197,29 @@ def id(self, value: str) -> None:
...
+        Tags setter.
+        It receives a list of tags and performs type validation
+        """
+        if not isinstance(values, list):
+            raise ValueError("Tags must be of type list")
+
+        for value in values:
+            if not isinstance(value, str):
+                raise ValueError(f"Each tag must be of type str: {value}")
+
```
aspacca, line 220: what about?
```
    self._tags = [value for value in values if isinstance(value, str)]
    if len(self._tags) != len(values):
        raise ValueError(f"Each tag must be of type str, given: {values}")
```

this way we print the original tags list

aspacca:
@bogdan-copocean can you add `tags` to [README-AWS.md](https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md)?

thanks

bogdan-copocean:
yes, I'll add two more tests in handler and I'll write the docs too

.gitignore
```diff
@@ -31,3 +31,4 @@ tests/elastic-serverless-forwarder-junit.xml
 *.code-workspace
 .pytest_cache/
 .python-version
+htmlcov/
```
aspacca, line 34: 👍

aspacca (approved):

PR #48, opened 2021-12-14
Stress test outcome

aspacca:
Enhancement

## What does this PR do?

Address performance issues emerged during stress testing the lambda

## Why is it important?
We reduce that change of failure and missing ingestions with a proper scaled cluster

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### v0.12.0 - 2021/12/14
 ##### Features
-* General performance refactoring as stress test outcome: [#48](https://github.com/elastic/elastic-serverless-forwarder/pull/48)
+* General performance refactoring after stress test outcome: [#48](https://github.com/elastic/elastic-serverless-forwarder/pull/48)
 
 ### v0.11.0 - 2021/11/04
 ##### Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v0.12.0 - 2021/12/14
+##### Features
+* General performance refactoring as stress test outcome: [#48](https://github.com/elastic/elastic-serverless-forwarder/pull/48)
+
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### v0.12.0 - 2021/12/14
+##### Features
+* General performance refactoring as stress test outcome: [#48](https://github.com/elastic/elastic-serverless-forwarder/pull/48)
```
endorama, line 3:
```suggestion
* General performance refactoring after stress test outcome: [#48](https://github.com/elastic/elastic-serverless-forwarder/pull/48)
```

endorama (changes requested):
Just a minor wording enhancement and a request: could you add to fafa0612c99330c0c1494913c3566f30af328a30 commit message some details about the change? I think they are very meaningful and merit to be added to the history for future reference.

aspacca:
> could you add to fafa061 commit message some details about the change?

reworded as https://github.com/elastic/elastic-serverless-forwarder/pull/48/commits/a0dfb687eeb21fad2aae8752e8c63591f928e451

aspacca:
@endorama fixed

endorama:
I've just a question around lambda RAM limits. Does it mean that if I have a file larger than max Lambda memory I cannot ingest it directly into Elastic, but it requires to be split in smaller chunks? Just for clarity, with the current 10GB RAM limit I don't this as a concern

endorama (approved):

aspacca:
> Does it mean that if I have a file larger than max Lambda memory I cannot ingest it directly into Elastic, but it requires to be split in smaller chunks?

that's correct, we could further improve and according to the length of the file given by the `head_object` request we can decide if using `get_object` with the eventual connection reset that we should find a way to handle, or `donwload_fileobj` in the current optimisation

Issue #54, opened 2021-12-15
SAR: Gracefull handling of error scenarios


ravikesarwani:
Handle error scenarios gracefully in the Lambda function. Few scenarios comes to mind that we need to investigate and see if extra error handling code needs to be added:

- [x] Elasticsearch overloaded “(429) Too Many Requests”
- [x] Elasticsearch unreachable because of network or other errors
- [x] Some events in the bulk request failed to ingest
- [x] Failure because of format errors etc.

Some error cases can be retried and we should have optimal retry logic in the code for those.
In other cases we will need to use the DLQ and write events there so users can investigate and potentially be able to re-process

aspacca:
We identified two kinds of failure that can happen during the execution of the Lambda:

1. Errors before the ingestion phase started
2. Errors during the ingestion phase

For errors at (1), we can safely make the Lambda return a failure: these errors are mostly due to misconfiguration, improper permission on the AWS resources, etc. Most importantly when an error occurs at this stage we don’t have any state yet about the events that are ingested, so there’s no consistency to keep and we can safely let the Lambda return a failure, with the underlying trigger ending up in a built-in retry mechanism.

For errors at (2) the situation is more tricky: we have now a state for N failed events out of total X events, if we fail the whole Lambda all the X events will be processed again. While the N failed ones could now succeed, the remaining X-N will now fail, since the datastreams are append-only and we would try to recreate already ingested documents (the ID of the document is deterministic).

The implemented solution to overcome this problem is to add an internal “reply queue”, similar to the internal continuing queue, where to store a single failed event per message that the users can optionally use to trigger the Lambda itself, or consume the messages in the way they prefer.
In the case that the reply queue is used to trigger the Lambda itself, a proper handler must be implemented.

In both cases, supporting a reply queue has the drawback of adding code to inspect every single failure in the Shippers (ie: output components in the ESF context) and handling the messaging to the reply queue.

While at the moment we support only Elasticsearch as output/shipper, the Lambda function is designed to easily support multiple outputs.

If this will be the case in the future, complexity of handling the messaging to the reply queue will grow linearly, since a single event could have different failure/success outcome for different outputs, and all of them have to be handled and sent to the reply queue separately.

[...]

PR #57, opened 2021-12-24
Introduce RaisableError exception and refactor wrap_try_except

aspacca:
Enhancement

## What does this PR do?
Introduces RaisableError exception and refactor wrap_try_except to raise it instead of gracefully logs and exit the lambda with success

## Why is it important?
See #54 
This are errors that can happen before the ingestion phase started
We can safely make the Lambda return a failure: these errors are mostly due to misconfiguration, improper permission on the AWS resources, etc

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

Relates #54

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -36,7 +36,7 @@ def capture_serverless(
     return apm_capture_serverless()(func)  # type:ignore
 
 
-class RaisebleException(Exception):
+class LambdaFailureableException(Exception):
     """Raised when the lambda must fail"""
 
     pass
@@ -47,14 +47,14 @@ def wrap_try_except(
 ) -> Callable[[dict[str, Any], context_.Context], str]:
     """
     Decorator to catch every exception and capture them by apm client if set
-    or raise if type is RaisebleException
+    or raise if type is LambdaFailureableException
     """
 
     def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> str:
         apm_client: Client = get_client()
         try:
             return func(lambda_event, lambda_context)
-        except RaisebleException as e:
+        except LambdaFailureableException as e:
             if apm_client:
                 apm_client.capture_exception()
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -36,8 +36,26 @@ def capture_serverless(
     return apm_capture_serverless()(func)  # type:ignore
 
 
-class LambdaFailureableException(Exception):
-    """Raised when the lambda must fail"""
+class TriggerTypeException(Exception):
+    """Raised when there is an error related to the trigger type"""
+
+    pass
+
+
+class ConfigFileException(Exception):
+    """Raised when there is an error related to the config file"""
+
+    pass
+
+
+class InputConfigException(Exception):
+    """Raised when there is an error related to the configured input"""
+
+    pass
+
+
+class OutputConfigException(Exception):
+    """Raised when there is an error related to the configured output"""
 
     pass
 
@@ -47,14 +65,15 @@ def wrap_try_except(
 ) -> Callable[[dict[str, Any], context_.Context], str]:
     """
     Decorator to catch every exception and capture them by apm client if set
-    or raise if type is LambdaFailureableException
+    or raise if type is of between TriggerTypeException, ConfigFileException,
+    InputConfigException or OutputConfigException
     """
 
     def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> str:
         apm_client: Client = get_client()
         try:
             return func(lambda_event, lambda_context)
-        except LambdaFailureableException as e:
+        except (ConfigFileException, InputConfigException, OutputConfigException, TriggerTypeException) as e:
             if apm_client:
                 apm_client.capture_exception()
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -81,6 +81,7 @@ def wrap_try_except(
 
             raise e
 
+        # NOTE: any generic exception is logged and suppressed to prevent the entire Lambda function to fail. As Lambas can process multiple events, when within a Lambda execution only some event produce an Exception it should not prevent all other events to be ingested.
         except Exception as e:
             if apm_client:
                 apm_client.capture_exception()
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -36,22 +36,38 @@ def capture_serverless(
     return apm_capture_serverless()(func)  # type:ignore
 
 
+class RaisebleException(Exception):
+    """Raised when the lambda must fail"""
+
+    pass
+
+
 def wrap_try_except(
     func: Callable[[dict[str, Any], context_.Context], str]
 ) -> Callable[[dict[str, Any], context_.Context], str]:
     """
     Decorator to catch every exception and capture them by apm client if set
+    or raise if type is RaisebleException
     """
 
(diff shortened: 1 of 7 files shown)


endorama (commented):
I have a concern about the `RaisableException` name. Given that all exceptions can be raised in Python, I think the naming is confusing. If I understood correctly, is a type of exception that always terminate lambda execution, while the code catches `Exception`, logs them and then suppress them.

In case this is correct, I would choose a more meaningful name (like `FatalException` or `UnrecoverableException`).

What do you think?

aspacca:
@endorama I can see your concern
I see `FatalException` or `UnrecoverableException` misleading the same

This is an exception that (if it occurs) we want to let the lambda fail since the next run could be solved (so it's not really unrecoverable) and doesn't let the lambda in a mixed state related to the ingestion phase that we cannot reconcile.

maybe something like `LambdaFailureableException`?

endorama:
I see the reason for "raisable" from your explanation. I proposed fatal or unrecoverable because this code is stateless, and from it's point of view it doesn't care if "the next run with clean state" will be able to perform the requested action. The current code execution is not able to progress and must halt. That's why I considered fatal a correct naming, as in the execution context this is not recoverable.

I see the point in not emitting something that appears to be an absolute failure that require human intervention, but I have some considerations about it:
- the standard expectation with Lambas is that they fail with unhandled exceptions; we are purposely not handling this exception as we want the lambda to fail, but we rely on a standard expectation for the environment the code is run in
- what happens when the same exception triggers multiple times? What 5 `LambdaFailureableException` tells to the operator?
- how is the operator expected to understand if it requires human intervention or not?
- "Raisable" "LambdaFailure" express more the intent of the developer than what the code is expected to do

So I would slightly change the implementation: instead of returning a generic exception that aims to "trigger lambda failure" I would raise different exceptions. Then in `wrapper` within `wrap_try_catch` I would handle them either suppressing them or allowing them to bubble up.
For example when `config_yaml` is empty the lambda must fail because there is no way execution can be successful. But is a different case from when config is missing output declaration or config cannot be parsed from SQS.

This would remove this naming issue and help operator react to specific occurrences.

aspacca:
>I see the point in not emitting something that appears to be an absolute failure that require human intervention, but I have some considerations about it:

it's rather the contrary indeed.
there are two kind of errors/exceptions that can happen:
1. before ingestion phase started
2. after ingestion phase started

for (1) we don't have any state yet, and we can let the lambda fail and be triggered again by the built-in retry mechanism.

>what happens when the same exception triggers multiple times? What 5 `LambdaFailureableException` tells to the operator?

this is related to (1), and it handled by the built-in retry mechanism according to users configuration: there could indeed be only one `LambdaFailureableException` or 5 according to the users setup, in the end all the failed "sources", after the max number of failures, could or could not go to a dead letter queue to be replayed or manually inspected. we don't have any control on this as is totally on the users's side.
that best we can do, from how I see it,  is to not ignore this type of error and let the lambda fail: then according to users' setup and needs they will decide how to handle the failure and how many times.
to clarify better: we should not silently ignore these failures like it is now since this will give no visibility to the users that something was wrong.

I agree we should be concerned about self-explaining naming of the error that will be reported

>how is the operator expected to understand if it requires human intervention or not?

That in my point of view should not be solely on the exception name, but rather on the exception message.
It could also be documented: the exception cases can be listed and for each of them a solution can be described.

>For example when config_yaml is empty the lambda must fail because there is no way execution can be successful. But is a different case from when config is missing output declaration or config cannot be parsed from SQS.

I don't see them as different cases, but for what the solution will be for each of them.
Every exception that aims to "trigger lambda failure" are the same: something that can have a fix the users have to apply.

This leads to errors of type (2):
the situation is more tricky here: we have now a state for N failed events out of total X events, if we fail the whole Lambda all the X events will be processed again. While the N failed ones could now succeed, the remaining X-N will now fail, since the datastreams are append-only and we would try to recreate already ingested documents (the ID of the document is deterministic).

We should not let the lambda fail here for this reason, since this will lead to a new failure (for different events) in the second execution run: a proposal to overcome this problem is to add an internal “reply queue”, similar to the internal continuing queue, where to store a single failed event per message that the users can optionally use to trigger the Lambda itself, or consume the messages in the way they prefer.

Basically the errors for (2) cannot and should not be fixed by the users but rather have a way to be handled (or at least their failed ingestion events stored to be inspected and eventually replayed) by the lambda itself.


>So I would slightly change the implementation: instead of returning a generic exception that aims to "trigger lambda failure" >I would raise different exceptions. Then in wrapper within wrap_try_catch I would handle them either suppressing them or allowing them to bubble up.
>This would remove this naming issue and help operator react to specific occurrences.

Instead of having a single `LambdaFailureableException` we could have `InvalidConfigException`, `InvalidTypeException` etc etc and bubble all of them in `wrap_try_catch` to "trigger lambda failure" 

I agree this would be more self explanatory, let's proceed this way

handlers/aws/utils.py
```diff
@@ -36,22 +36,57 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
             return func(lambda_event, lambda_context)
+        except (ConfigFileException, InputConfigException, OutputConfigException, TriggerTypeException) as e:
+            if apm_client:
+                apm_client.capture_exception()
+
+            shared_logger.exception("exception raised", exc_info=e)
+
+            raise e
+
         except Exception as e:
```
endorama, line 84:
```suggestion
        # NOTE: any generic exception is logged and suppressed to prevent the entire Lambda function to fail. As Lambas can process multiple events, when within a Lambda execution only some event produce an Exception it should not prevent all other events to be ingested.
        except Exception as e:
```

handlers/aws/utils.py
```diff
@@ -36,22 +36,57 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
     Decorator to catch every exception and capture them by apm client if set
+    or raise if type is of between TriggerTypeException, ConfigFileException,
+    InputConfigException or OutputConfigException
     """
 
     def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> str:
+        apm_client: Client = get_client()
         try:
             return func(lambda_event, lambda_context)
+        except (ConfigFileException, InputConfigException, OutputConfigException, TriggerTypeException) as e:
```
endorama, line 76:
```suggestion
        # NOTE: for all these cases we want the exception to bubble up to Lambda platform and let the defined retry mechanism take action. These are non transient unrecoverable error from this code point of view.
        except (ConfigFileException, InputConfigException, OutputConfigException, TriggerTypeException) as e:
```

endorama (approved):
Added a couple of comments to clarify the expected behaviour of exceptions per the discussion in https://github.com/elastic/elastic-serverless-forwarder/pull/57#issuecomment-1002123112 but 👍 from me

endorama:
> This leads to errors of type (2):
the situation is more tricky here: we have now a state for N failed events out of total X events, if we fail the whole Lambda all the X events will be processed again. While the N failed ones could now succeed, the remaining X-N will now fail, since the datastreams are append-only and we would try to recreate already ingested documents (the ID of the document is deterministic).
> We should not let the lambda fail here for this reason, since this will lead to a new failure (for different events) in the second execution run: a proposal to overcome this problem is to add an internal “reply queue”, similar to the internal continuing queue, where to store a single failed event per message that the users can optionally use to trigger the Lambda itself, or consume the messages in the way they prefer.

I was wondering if this failure mode should be discussed separately, as I see this both as interesting from our customers perspective and as insightful for contributors. Are we tracking this discussion elsewhere already?

Issue #59, opened 2021-12-30
Add a NOTICE.txt file


masci:
A file to track the 3rd party licenses used in this project should be added at the root of this repo

PR #61, opened 2022-01-11
notice file

bogdan-copocean:
- Docs

## What does this PR do?
Adds NOTICE.txt file to the project

## Why is it important?
See #59 

## Checklist
- [] My code follows the style guidelines of this project
- [] I have commented my code, particularly in hard-to-understand areas
- [] I have made corresponding changes to the documentation
- [] I have made corresponding change to the default configuration files
- [] I have added tests that prove my fix is effective or that my feature works
- [] I have added an entry in `CHANGELOG.md`

## Author's Checklist

- [ ]

## Related issues
- Relates #59


tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+        existing_packages.sort()
+
+        requirements_name_from_file = [requirement for requirement in self.required_packages.keys()]
+        requirements_name_from_file.sort()
+
+        real_requirements_name = [requirement for requirement in self.required_packages.values()]
+        real_requirements_name.sort()
+
+        if self.mode == "--check":
+            if real_requirements_name == existing_packages:
```
aspacca, line 55: repeated at https://github.com/elastic/elastic-serverless-forwarder/pull/61/files#diff-68dab1ed5621e764af621e1ab471747e31c7615e6f8cf47039e6509d6fd6336fR68-R71

we can just have it once outside of the if/else on `self.mode`

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+        and self.POSSIBLE_METADATA_FILES where the important information about the package should exist
+        """
+        for entry in self.scanned_results_json["files"]:
+            splitted_entry_path: list[str] = entry["path"].split("/")
+
+            if (
+                splitted_entry_path[-1] in NoticeParser.POSSIBLE_LICENSE_FILES
+                or splitted_entry_path[-1] in NoticeParser.POSSIBLE_METADATA_FILES
+            ):
+                if not len(splitted_entry_path) > 6:
```
aspacca, line 118: what's this for?

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+                    continue
+
+                package_name = splitted_entry_path[5].split("-")[0]
```
aspacca, line 121: what's at offset `5`?

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+
+                if package_name == required_package:
+                    if package_name not in self.processed_packages:
+                        # print(entry)
```
aspacca, line 125: please remove

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+                            self.processed_packages[package_name]["license_name"] = entry["licenses"][0]["key"].upper()
+
+                    if splitted_entry_path[-1] in NoticeParser.POSSIBLE_METADATA_FILES:
+                        package_version: str = splitted_entry_path[5].split("-")[1].strip(".dist")
+                        self.processed_packages[package_name]["version"] = package_version
+
+                        homepage_url: str = entry["packages"][0]["homepage_url"]
+                        vcs_url: str = entry["packages"][0]["vcs_url"]
+
+                        if not homepage_url and not vcs_url:
```
aspacca, line 139: can we have a license but not the `Homepage URL` or `VCS URL`?
In this case we should not skip the entry but just omit the `homepage_url` value

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+        if "license_content" not in self.processed_packages[processed_package]:
+            try:
+                raw_github_base_url: str = "https://raw.githubusercontent.com"
+                homepage_url: str = self.processed_packages[processed_package]["homepage_url"]
+                possible_github_project: str = ""
+
+                if "github" in homepage_url:
+                    possible_github_project = "/".join(homepage_url.split("/")[3:])
+
+                github_license_pages: list[str] = [
```
aspacca, line 225: 👍

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+
+                    if response.status_code == 200:
+                        self.processed_packages[processed_package]["license_content"] = response.text
+                        self.processed_packages[processed_package]["license_path"] = github_page
+                        break
+                    else:
+                        print(f"[-] License could not be found at: {github_page}")
+
+            except Exception as e:
+                print(f"[-] {e}")
```
aspacca, line 247: debug leftover?

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+from requests import Response
+
+
+class NoticeParser:
+    POSSIBLE_LICENSE_FILES: list[str] = [
+        "LICENSE",
+        "LICENSE.txt",
+        "LICENSE.rst",
+        "apache-2.0.LICENSE",
+        "apache-2.0.LICENSE.txt",
```
aspacca, line 21: this list seems a little short, or is it only my impression? :)

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+
+                    if not processed_package:
+                        print(
+                            f"[-] Nothing has been found for package '{real_package_name}' in {self.scanned_json_file}",
+                        )
+                        continue
+
+                    if (
+                        "package_name" not in processed_package
+                        or "version" not in processed_package
```
aspacca, line 90: not sure we should skip if missing `version`

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+                        or "homepage_url" not in processed_package
```
aspacca, line 91: I think we should not skip if missing `homepage_url`

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+                        or "license_name" not in processed_package
+                        or "license_path" not in processed_package
```
aspacca, line 93: I think we should not skip if missing `license_path`

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+                        or "license_content" not in processed_package
```
aspacca, line 94: for `license_content` we could add a lookup on https://spdx.org/licenses/ as extra step
 
not sure if, once we identify the license name, missing the content we should skip the entry at all.

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+            else:
+                raise
+
+        except Exception as e:
+            print(f"[-] {e}")
+            raise
+        else:
+            return license_content
+
+    def read_requirements(self) -> None:
```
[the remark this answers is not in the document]
bogdan-copocean: I'm not sure about that.. from what I see, the requirements files usually come in `aws_lambda_typing>=2,<2.1`,  `localstack[runtime]>=0.12.19,<0.13`, `docker==5.0.0` or simply `isort`

tests/scripts/parse_notice.py
```diff
@@ -0,0 +1,277 @@
...
+
+                with open("NOTICE.txt") as fh:
+                    license_content = fh.read()
+
+                return license_content
+            else:
+                raise
+
+        except Exception as e:
+            print(f"[-] {e}")
```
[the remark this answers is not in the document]
bogdan-copocean: I tried to prefix errors with [-] and info/success with [!], but I'll remove those. It doesn't look that good

tests/scripts/parse_notice.py
bogdan-copocean: this is found in the `entry["path"]` key: `elastic-serverless-forwarder/venv/lib/python3.9/site-packages/pytest-6.2.5.dist-info/LICENSE`, and I split the path in order to access the name of the package and the LICENSE/METADATA file `['elastic-serverless-forwarder', 'venv', 'lib', 'python3.9', 'site-packages', 'pytest-6.2.5.dist-info', 'LICENSE']`.

That's what I'm checking there, if the LICENSE/METADATA exist in the possible file names and if the length of the splitted is > 6 because it can split paths like `elastic-serverless-forwarder/LICENSE.txt`.

tests/scripts/parse_notice.py
bogdan-copocean: `['elastic-serverless-forwarder', 'venv', 'lib', 'python3.9', 'site-packages', 'pytest-6.2.5.dist-info', 'LICENSE']` offset `5`: `pytest-6.2.5.dist-info` => package_name = `pytest`

tests/scripts/parse_notice.py
bogdan-copocean: Usually doesn't matter what type of license it is (MIT, apache, etc.) they're usually found in `LICENSE.txt` or `LICENSE`, but people may use other file names for those like "apache-2.0.License" etc.
These are the files I've encountered so far, but I'll add more possible combinations.

tests/scripts/parse_notice.py
bogdan-copocean: i checked here because I'm accessing those keys in the `write_to_file` function and it would run into a `KeyError`, and I didn't want to check in the `write_to_file` function. 

I'll refactor, and maybe add the keys with empty strings as values?

tests/scripts/parse_notice.py
bogdan-copocean: same as above

tests/scripts/parse_notice.py
bogdan-copocean: same as above

tests/scripts/parse_notice.py
bogdan-copocean: same as above. 

What would happen if we don't have anything for the license content for that package? Shall we add a general license based on the license name?

tests/scripts/scancode_plugins/src/notice_generator/run_notice_generator.py
```diff
@@ -0,0 +1,61 @@
...
+        check_notice, fix_notice = self.is_enabled(check_notice, fix_notice)
+
+        if check_notice:
+            mode = "check"
+        elif fix_notice:
+            mode = "fix"
+        else:
+            return
+
+        requirements_list: list[str] = ["requirements.txt", "requirements-lint.txt", "requirements-tests.txt"]
```
aspacca, line 53: the values in the list should not be hardcoded

tests/scripts/scancode_plugins/src/notice_generator/notice_generator.py
```diff
@@ -0,0 +1,287 @@
...
+        if not package_homepage_url:
+            package_homepage_url = ""
+
+        if not package_license_path:
+            package_license_path = ""
+
+        if not package_license_content:
+            package_license_content = ""
+
+        with open(self.notice_file_name, "a+") as fh:
```
aspacca, line 251: this should be better to move to a default scancode template/output on the long run

PR #62, opened 2022-01-13
README-AWS refinition

aspacca:
Docs
## What does this PR do?
Refine the docs for AWS

## Why is it important?
This content is presented to the users in the ESF page on SAR: providing the best curated content will help the users

## Checklist

~- [ ] My code follows the style guidelines of this project~
~- [ ] I have added an entry in `CHANGELOG.md`~

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/docs/README-AWS.md
+++ b/docs/README-AWS.md
@@ -55,7 +55,7 @@ At a high level the deployment consists of the following steps:
   * Still in the "Configuration" tab select "Permissions"
     * Click on the link of the IAM role for the Lambda under *Execution role* -> *Role name*
     * In the new window add a new policy to the role, as described at [Lambda IAM permissions and policies](#lambda-iam-permissions-and-policies)
-  * Back to the the "Configuration" tab in the Lambda window select "Triggers"
+  * Back to the "Configuration" tab in the Lambda window select "Triggers"
     * You can see an already defined SQS trigger for a queue with the prefix `elastic-serverless-forwarder-continuing-queue-`. This is an internal queue and should not be modified, disabled or removed.
     * Click on "Add trigger"
     * From "Trigger configuration" dropdown select "SQS"
--- a/docs/README-AWS.md
+++ b/docs/README-AWS.md
@@ -8,7 +8,7 @@ The Elastic Serverless Forwarder is an AWS Lambda function that ships logs from 
 
 The config yaml file (details described below) acts as an input where the user, based on input type, configures things like SQS queue ARN and Elasticsearch connection information. Multiple input sections can be created in the configuration file pointing to different queues that match specific log types.
 
-A continuing SQS queue is set up by the Lambda deployment automatically. It is used to trigger a new function invocation so that Lambda can continue from exactly where the last function run was terminate. By default a Lambda function runs for a max of 15 minutes. When processing large log files there’s a possibility that the function may be exited by AWS in the middle of processing a log file. The code handles this scenario gracefully by keeping track of the file and offset its processing.
+A continuing SQS queue is set up by the Lambda deployment automatically. It is used to trigger a new function invocation so that Lambda can continue from exactly where the last function run was terminated. By default a Lambda function runs for a max of 15 minutes. When processing large log files there’s a possibility that the function may be exited by AWS in the middle of processing a log file. The code handles this scenario gracefully by keeping track of the file and offset its processing.
 
 Application supports automatic routing of various AWS service logs to the corresponding data streams for further processing and storage in the Elasticsearch cluster.
 
@@ -26,6 +26,7 @@ The dataset parameters in the config file are optional. Lambda supports automati
 At a high level the deployment consists of the following steps:
 
 **Step1:** Install appropriate integration(s) from the Kibana UI. This sets up appropriate pre-built dashboards, ingest node configurations, and other assets that help you get the most out of the data you ingest. To see the full list of available integrations and install appropriate integrations, go to **Management** > **Integrations** in Kibana UI. For example browse or search for AWS, select the AWS integration, select Settings and click Install AWS assets to install all the AWS integrations.
+
 **Step2:** Browse for elastic-serverless-forwarder in SAR repository in AWS Console. Review and deploy the application and provide appropriate configuration information.
 
 ## How to deploy Elastic Serverless Forwarder Lambda application from the AWS Serverless Application Repository.
@@ -48,7 +49,6 @@ On top of this basic permission the following policies must be provided:
 * For every SQS queue resource that used as triggers of the Lambda the following action must be allowed:
   * `sqs:GetQueueUrl`
 
-
 * For every S3 bucket resource that SQS queues are receiving notification from used by triggers of the Lambda the following action must be allowed on the S3 buckets' keys:
   * `s3:GetObject`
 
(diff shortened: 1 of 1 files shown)


docs/README-AWS.md
```diff
@@ -2,68 +2,53 @@
...
+
 
 ### Deployment:
 At a high level the deployment consists of the following steps:
-**Step1:** Install appropriate integration(s) from the Kibana UI. This sets up appropriate pre-built dashboards, ingest node configurations, and other assets that help you get the most out of the data you ingest. To see the full list of available integrations and install appropriate integrations, go to **Management** > **Integrations** in Kibana.
-**Step2:** Browse for elastic-serverless-forwarder in SAR repository in AWS Console. Review and deploy the application and provide appropriate configuration information. Below are the detailed steps.
```
ravikesarwani, line 27: Add a space so Step2 shows in a separate line.

docs/README-AWS.md
```diff
@@ -2,68 +2,53 @@
...
   * `sqs:SendMessage`
 
 * For every S3 bucket resource that's reported in the `S3_CONFIG_FILE` environment variable the following action must be allowed on the S3 buckets' config file object key:
   * `s3:GetObject`
 
+* For every SQS queue resource that used as triggers of the Lambda the following action must be allowed:
+  * `sqs:GetQueueUrl`
+
+
 * For every S3 bucket resource that SQS queues are receiving notification from used by triggers of the Lambda the following action must be allowed on the S3 buckets' keys:
```
ravikesarwani, line 52: I though the above was setup automatically when the user adds the trigger in the UI.
I think we should mention it as a dependency but was thinking that it maybe good to call out more as a verification step rather than something that they need to setup.

docs/README-AWS.md
```diff
@@ -75,35 +60,7 @@ On top of this basic permission the following policies must be provided:
 
 * For SQS queue resource that you want to use as triggers of the Lambda the proper permission are already included by `arn:aws:iam::aws:policy/service-role/AWSLambdaSQSQueueExecutionRole`:
```
ravikesarwani, line 62: Either this is true or line 48. Can we combine the IAM requirement for trigger SQS queue in one place.

aspacca: Their are both true someway.

Line 48 is related to specific error handling in batch processing of sqs messages.
The most straightforward and reliable way to handle them is to delete a message once it's processed, this lead to the needs of `GetQueueUrl` permission

Another option is what's describe here: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting

In my opinion is much more fragile and requires extra configuration step for the users on their own queues.

docs/README-AWS.md
```diff
@@ -127,7 +84,7 @@ Resources:
       }
     },
     {
-      "PolicyName": "ElasticServerlessForwarderFunctionRolePolicySQSContinuingQueue", ## ADD AS IT IS FOR THE REPLAY QUEUE
+      "PolicyName": "ElasticServerlessForwarderFunctionRolePolicySQSReplayQueue", ## ADD AS IT IS FOR THE REPLAY QUEUE
       "PolicyDocument": {
```
ravikesarwani, line 88: I was wondering if "Fn:GetAtt" on "ElasticServerlessForwarderContinuingQueue" will work in all enviornments?
For instance in my enviornment the queue name is "elastic-serverless-forwarder-continuing-queue-0ee8749bc847". Maybe I don't understand this fully but can we double check please?

In my setup I added a policy via copying ARN.

docs/README-AWS.md
aspacca: We might/should rephrase line 61 probably. Since it doesn't reflect 100% what's required

docs/README-AWS.md
aspacca: You are right, `Fn:GetAtt` is specific to cloudformation template.
I will split this two policies for cloudformation and console

docs/README-AWS.md
```diff
@@ -161,6 +118,24 @@ Resources:
...
+            ...
+          ],
+          "Effect": "Allow"
+          }
+        ]
+      }
+    },
     {
       "PolicyName": "ElasticServerlessForwarderFunctionRolePolicyS3", ## ADD FOR YOUR S3 BUCKET
       "PolicyDocument": {
```
ravikesarwani, line 141: The policy name here is same as for SQS queue from before.

docs/README-AWS.md
```diff
@@ -218,6 +193,64 @@ Resources:
...
+* Go to "Lambda > Functions" page in the AWS console and look for the Function Name with prefix "serverlessrepo-elastic-se-ElasticServerlessForward-" and click on it
+  * Go to "Configuration" tab and select "Environment Variables"
+  * Click on "Edit" on the "Environment Variables" tab and add the environment variable `S3_CONFIG_FILE` with the value of the S3 url in the format "s3://bucket-name/config-file-name" point to the configuration file for your Elastic Forwarder for Serverless (see below)
+  * You can additionally add the following environment variables to enable Elastic APM instrumentation to your deployment of Elastic Forwarder for Serverless
+    * | Key                       | Value  |
+      |---------------------------|--------|
+      |`ELASTIC_APM_ACTIVE`       | `true` |
+      |`ELASTIC_APM_SECRET_TOKEN` | token  |
+      |`ELASTIC_APM_SERVER_URL`	  | url    |
+  * Still in the "Configuration" tab select "Triggers"
```
ravikesarwani, line 217: I feel the section "Lambda IAM permissions and policies" should be just above section "setting the trigger" so that the flow seems right.

ravikesarwani (commented):
Other comments (S3_CONFIG_FILE section):
In the "S3_CONFIG_FILE" section there's a type "apy_key takes precedence".
Fix this line "inputs.[].type: The type of the trigger input (currently only sqs supported)". Change to "s3-sqs".
Add in line [args.dataset: Dataset for the data stream where to forward the logs to. Default value: "generic"] that this is optional. Lambda supports automatic routing of various AWS service logs to the corresponding data streams for further processing and storage in the Elasticsearch cluster. It supports automatic routing of aws.cloudtrail, aws.cloudwatch_logs, aws.elb_logs, aws.firewall_logs, aws.lambda, aws.sns, aws.s3_storage_lens, aws.vpcflow, and aws.waf logs. For other log types the users can optionally set the dataset value in the configuration file. If the dataset is not specified and it cannot be matched with any of the above AWS services then the dataset will be set to "generic".

ravikesarwani:
In the "Secrets Manager Support" section the config file example has line [id: "arn:aws:secretsmanager:eu-central-1:123-456-789:secret:plain_text_secret"]. Is this intentional? This will point to the SQS queue for trigger.
Can you clarify please?

docs/README-AWS.md
aspacca: not sure about I understand how to fix this
the "Lambda IAM permissions and policies" section is quite big, "setting the trigger" is just a bullet point

aspacca:
> In the "Secrets Manager Support" section the config file example has line [id: "arn:aws:secretsmanager:eu-central-1:123-456-789㊙️plain_text_secret"]. Is this intentional? This will point to the SQS queue for trigger. Can you clarify please?

yes, in the section there are examples for both plain text secret and secret with json content.
the difference is explained below in "Usage" section

PR #66, opened 2022-02-07
Kinesis data stream input

### LINKED ISSUE #53: SAR: Kinesis data streams input support
Add support for Kinesis data streams input

This helps the use case where users are already using Kinesis data streams to consolidate events from various services.
If they are already using Kinesis data streams we want to integrate without having them to create extra AWS setup.

aspacca:
Enhancement
## What does this PR do?
Add support for kinesis data stream as input

## Why is it important?
It gives the users more flexibility on the source of their logs

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

Closes #53

### DIFF
--- a/handlers/aws/kinesis_trigger.py
+++ b/handlers/aws/kinesis_trigger.py
@@ -29,6 +29,7 @@ def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dic
 
     events = storage.get_by_lines(range_start=0)
 
+    ignored_value = -1
     for log_event, ending_offset, newline_length in events:
         assert isinstance(log_event, bytes)
 
@@ -49,4 +50,4 @@ def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dic
 
         es_event["fields"]["cloud"]["region"] = aws_region
 
-        yield es_event, ending_offset, -1, -1
+        yield es_event, ending_offset, ignored_value, ignored_value
--- a/handlers/aws/kinesis_trigger.py
+++ b/handlers/aws/kinesis_trigger.py
@@ -13,7 +13,7 @@ from .event import _default_event
 from .utils import get_kinesis_stream_name_type_and_region_from_arn
 
 
-def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dict[str, Any], int, int, int]]:
+def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dict[str, Any], int]]:
     """
     Handler for kinesis data stream inputs.
     It iterates through kinesis records in the kinesis trigger and process
@@ -29,7 +29,6 @@ def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dic
 
     events = storage.get_by_lines(range_start=0)
 
-    ignored_value = -1
     for log_event, ending_offset, newline_length in events:
         assert isinstance(log_event, bytes)
 
@@ -50,4 +49,4 @@ def _handle_kinesis_record(kinesis_record: dict[str, Any]) -> Iterator[tuple[dic
 
         es_event["fields"]["cloud"]["region"] = aws_region
 
-        yield es_event, ending_offset, ignored_value, ignored_value
+        yield es_event, ending_offset
--- a/handlers/aws/kinesis_trigger.py
+++ b/handlers/aws/kinesis_trigger.py
@@ -1,0 +1,52 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+import datetime
+from copy import deepcopy
+from typing import Any, Iterator
+
+from share import shared_logger
+from storage import CommonStorage, StorageFactory
+
+from .event import _default_event
+from .utils import get_kinesis_stream_name_type_and_region_from_arn
+
+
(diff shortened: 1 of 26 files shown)


aspacca:
cannot reproduce test flakiness locally

tests/handlers/aws/test_handler.py
```diff
@@ -545,8 +546,387 @@ def test_lambda_handler_failure(self) -> None:
...
+
+    client.create_bucket(Bucket=bucket_name, ACL="public-read-write")
+    client.put_object(Bucket=bucket_name, Key=key_name, Body=content, ContentType=content_type)
+
+
 @pytest.mark.integration
-class TestLambdaHandlerSuccess(TestCase):
+class TestLambdaHandlerSuccessKinesisDataStream(TestCase):
+    def setUp(self) -> None:
+        docker_client = docker.from_env()
```
endorama, line 584: Just wanted to highlight [`testcontainers-python`](https://testcontainers-python.readthedocs.io/en/latest/index.html). Have you considered using it instead of plain docker? Could it help reducing boilerplate?

aspacca: nice! I didn't know about that.
I have to investigate the feature of the provided `LocalStack` container.
the boilerplate is mostly about getting the random port opened by docker on the host and override it for the `localstack` utils package.

I will open an issue in order to check how much is feasible

aspacca: https://github.com/elastic/elastic-serverless-forwarder/issues/67

aspacca:
@v1v the integration tests went flaky for the PR.
my guess is that it's due to resource contention in the CI workers, since I'm not able to reproduce locally both running tests directly on the host or inside docker.
do you have any hint or workaround (like increasing the workers size, at least to assess this is really the issue)?

thanks

v1v:
> do you have any hint or workaround (like increasing the workers size, at least to assess this is really the issue)?

As far as I know the workers size is big for beats-ci. `n1-highmem-8` type (`8 vCPU | RAM 53248 MiB | Storage 10Gb Standard (Non-SSD)`

My suggestion will consist on booking one of the running CI workers and ssh onto, to run the same steps there.

There is an option to use a bigger machine dynamically within the pipeline, but I'd say it might better to iterate with debugging within the CI worker context, I'll send you those details privately.

handlers/aws/kinesis_trigger.py
```diff
@@ -0,0 +1,52 @@
...
+            "kinesis": {
+                "type": stream_type,
+                "name": stream_name,
+                "sequence_number": kinesis_record["kinesis"]["sequenceNumber"],
+            }
+        }
+
+        es_event["fields"]["cloud"]["region"] = aws_region
+
+        yield es_event, ending_offset, -1, -1
```
endorama, line 52: Is it possible to name these magic numbers? Their meaning is quite obscure

endorama (commented):
PR is huge, so I may have missed something. Just a minor nitpick but seems ok to me.

handlers/aws/kinesis_trigger.py
aspacca: more than magic they are useless 😭 
they are ignored by the receiver (https://github.com/elastic/elastic-serverless-forwarder/pull/66/files#diff-d2b852c7ba6cf83d5e104cfb82ab3e418afc8f060dc370acdac7da58a4cac902R110)

aspacca:
>There is an option to use a bigger machine dynamically within the pipeline, but I'd say it might better to iterate with debugging within the CI worker context, I'll send you those details privately.

I will just use a bigger machine once to assess if it's really a resource issue. if not I will ssh to the CI worker and debug

aspacca:
>PR is huge, so I may have missed something. Just a minor nitpick but seems ok to me.

sorry, I tried to split the commits to be the more atomically possible, please let me know any improvement I can apply for easing the review :)

handlers/aws/kinesis_trigger.py
```diff
@@ -0,0 +1,53 @@
...
+
+    events = storage.get_by_lines(range_start=0)
+
+    ignored_value = -1
+    for log_event, ending_offset, newline_length in events:
+        assert isinstance(log_event, bytes)
+
+        es_event = deepcopy(_default_event)
+        es_event["@timestamp"] = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+        es_event["fields"]["message"] = log_event.decode("UTF-8")
```
kaiyan-sheng, line 38: I think from the ARN, you can also get the `cloud.account.id` and store it in a separate field.

handlers/aws/kinesis_trigger.py
aspacca: silly me! I don't need to have the same signature of other handles :)

handlers/aws/kinesis_trigger.py
aspacca: I will open a separate ticket for it, reviewing all metadata enrichment

aspacca: https://github.com/elastic/elastic-serverless-forwarder/issues/68

aspacca:
@v1v 

>I will just use a bigger machine once to assess if it's really a resource issue. if not I will ssh to the CI worker and debug

I started with a `n1-highmem-16` and got a green, but soon after still the same recurring flakiness.
I've ssh'ed to the worker and found rather low memory usage but quite high cpu one.
I switched so to `c2-standard-8` that's compute optimised, and it seems stable now (it's even a 12% cheaper).

there is still some known flakiness that I am able to hit locally and I will fix in an another PR

handlers/aws/kinesis_trigger.py
kaiyan-sheng: Sounds good!

kaiyan-sheng (approved):
Looks good to me!

PR #69, opened 2022-02-09
Multiple tweaks

aspacca:
Cleanup

## What does this PR do?
ES client:
- sets max retries to 4: max value according to `initial_backoff * 2 ** retry_number` with `initial_backoff = 2` and max backoff 600 secs
- enables http compress by default

Internal renaming of `trigger_type` for replay queue to `replay-sqs`

## Why is it important?
Cleaner code

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v0.20.1 - 2022/02/08
+##### Bug fixes
+* HTTP compression always on and max retries to not exceed 15 mins in the ES client [#68](https://github.com/elastic/elastic-serverless-forwarder/pull/68)
+
 ### v0.20.0 - 2022/02/07
 ##### Features
 * Add support for `kinesis-data-stream` input: [#66](https://github.com/elastic/elastic-serverless-forwarder/pull/66)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### v0.20.1 - 2022/02/08
+##### Bug fixes
+* HTTP compression always on and max retries to not exceed 15 mins in the ES client [#69](https://github.com/elastic/elastic-serverless-forwarder/pull/69)
```
masci, line 3: I feel like a verb is missing, maybe you meant `Set HTTP compression...`?

endorama (approved):
One question, but is not a blocker: why limiting retry to 15 minutes? Is this mandated by something on ES side?

aspacca:
> One question, but is not a blocker: why limiting retry to 15 minutes? Is this mandated by something on ES side?

the lambda cannot run more than 15 minutes. assuming a single bulk request will keep failing we should still be able to return before the lambda times out so that it can execute the error handling code

CHANGELOG.md
aspacca: indeed: explicit is better than implicit :)

PR #73, opened 2022-02-14
Allow es sending to non datastream

### LINKED ISSUE #72: Improve handling of index/data stream routing
Provide a single configuration parameter (ex: es_index_or_datastream_name) for users to **optionally** set the index or data stream name where the data is sent to Elasticsearch. The users should be able to use an alias as well.
As part of this change we should remove the current **dataset** and **namespace** config values in lieu of this single config value.

For various supported AWS service logs the user doesn’t have to set es_index_or_datastream_name as the code will automatically process the supported AWS services log and send it to the correct integration data stream. This assumes that user has installed the AWS integration assets.
If there is no value specified by the user and also we do not have a match against any of the  supported AWS service logs then the data should be sent to "logs-generic-default" data stream.

If es_index_or_datastream_name is specified by the user in the input section of the config, the Lambda will parse the value and check if it matches to the convention of an integration data stream `<type>-<dataset>-<namespace>`. If there is a pattern match, it will process it as an integration data stream.
If there is no pattern match then the code will process it as a regular **index** and send the data without adding any data streams specific parameters.

The above logic doesn’t handle the case of user defined custom data stream name that doesn’t follow the integration data stream naming convention `<type>-<dataset>-<namespace>`. We will handle that use case as a separate issue.

- [x] Update code
- [x] Update documentation 

cc: @aspacca

aspacca:
Breaking change

## What does this PR do?
Add support for both index and datstream for sending events do elasticsearch

## Why is it important?
Not every users might use datastreams, and they need a way to send to an index or an alias: before this change it was not possible.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-
Closes #72


docs/README-AWS.md
```diff
@@ -469,8 +467,7 @@ Custom init arguments for the given forwarding target output
...
-  * `args.namespace`: Namespace for the data stream where to forward the logs to. Default value: "default"
+  * `args.es_index_or_datastream_name`: Name of the index or data stream where to forward the logs to. Lambda supports automatic routing of various AWS service logs to the corresponding data streams for further processing and storage in the Elasticsearch cluster. It supports automatic routing of `aws.cloudtrail`, `aws.cloudwatch_logs`, `aws.elb_logs`, `aws.firewall_logs`, `aws.lambda`, `aws.sns`, `aws.s3_storage_lens`, `aws.vpcflow`, and `aws.waf` logs. For other log types, if using data stream, the users can optionally set its value in the configuration file according to the naming convention for data streams and available integrations. If the           `es_index_or_datastream_name` is not specified and it cannot be matched with any of the above AWS services then the value will be set to "logs-generic-default".
```
endorama, line 470:
```suggestion
  * `args.es_index_or_datastream_name`: Name of the index or data stream where to forward the logs to. Lambda supports automatic routing of various AWS service logs to the corresponding data streams for further processing and storage in the Elasticsearch cluster. It supports automatic routing of `aws.cloudtrail`, `aws.cloudwatch_logs`, `aws.elb_logs`, `aws.firewall_logs`, `aws.lambda`, `aws.sns`, `aws.s3_storage_lens`, `aws.vpcflow`, and `aws.waf` logs. For other log types, if using data stream, the users can optionally set its value in the configuration file according to the naming convention for data streams and available integrations. If the `es_index_or_datastream_name` is not specified and it cannot be matched with any of the above AWS services then the value will be set to "logs-generic-default".
```

endorama (commented):
I'm ok with the change but I would like to discuss the implementation. I think a different option (like `index_name`) would be preferred over the composite `es_index_or_datastream_name`. If set, it overrides data stream naming, otherwise we retain the same behaviour.

This would have a set of advantages:
- not a breaking change (old config would still work)
- allow to configure the index name if needed
- in case data streams are used the current config reduce cognitive load for customers (as they have to specify only data set and namespace and not worry about data stream naming constraints).
- clearer configuration logic: either data streams or override (override supports writing to indexes too)

What to you think?

aspacca:
> I'm ok with the change but I would like to discuss the implementation. I think a different option (like `index_name`) would be preferred over the composite `es_index_or_datastream_name`. If set, it overrides data stream naming, otherwise we retain the same behaviour.
> 
> This would have a set of advantages:
> 
> * not a breaking change (old config would still work)
> * allow to configure the index name if needed
> * in case data streams are used the current config reduce cognitive load for customers (as they have to specify only data set and namespace and not worry about data stream naming constraints).
> * clearer configuration logic: either data streams or override (override supports writing to indexes too)
> 
> What to you think?

I discussed the option of an extra config param without having a breaking change with @ravikesarwani 

the golden rule we agreed on was "let's make it work with defaults and/or plain config for 90% of the use cases"
that's the reason for a single config
yes, the change will be breaking, being in beta and still having users in the POC phase and not in the production one we decided it is fine to do.
regarding the override and more granular configuration (in this PR you cannot override the single value of the `namespace` for example), sticking to the golden rule, in a next iteration, an extra/overriding way for configuring the 10% of the use cases that don't fit will be provided. this will properly documented in a separated section of the docs as something like `advanced config`, for users that need to tweak the setup to their very specific cases

do you think that's reasonable?

@ravikesarwani , do you think we should reconsider the implementation according to @endorama suggestions?

aspacca:
@endorama I forgot to link the issue https://github.com/elastic/elastic-serverless-forwarder/issues/72

ravikesarwani:
> in case data streams are used the current config reduce cognitive load for customers (as they have to specify only data set and namespace and not worry about data stream naming constraints).

While in theory this may seem like it, in practical (and based on our experience talking to folks in GoDaddy) no one really knows what "dataset" & "namespace" really is (even after giving them all the links for data stream naming conventions) and how we internally create the full data stream name by combining "logs-", dataset & namespace.

Having a single value is desirable and many Elasticsearch users understand index or data stream name. This is still an optional value (for AWS service logs we send it to correct data streams and they don't have to specify this) and they will only use this config when they have created their own index template (in complex use cases) and its easier for them to put the full value or the alias here. Breaking it into many pieces is our internal logic and not something I think we need to expose to our customers via way of multiple config values.

endorama (approved):

endorama:
> While in theory this may seem like it, in practical (and based on our experience talking to folks in GoDaddy) no one really knows what "dataset" & "namespace" really is (even after giving them all the links for data stream naming conventions) and how we internally create the full data stream name by combining "logs-", dataset & namespace.
> 
> Having a single value is desirable and many Elasticsearch users understand index or data stream name. 

@ravikesarwani this is a really interesting comment, thanks for sharing.

Approved, code was already :heavy_check_mark:

PR #95, opened 2022-03-15
Fix handling of different event sources

aspacca:
Enhancement
## What does this PR do?
Fix handling of batch messages in the continuing queue where the original event sources are different for every message when the batch size is greater than 1

## Why is it important?
The continuing queue receive messages from different original inputs: `cloudwatch-logs`, `sqs`, `s3-sqs`
When we started the development this complexity was not taken in consideration, and the type and id of the input from the first record of the queue was used. While this behaviour is reliable for user provided event inputs it breaks the correctness of the processing when applied to the continuing queue, where the same batch can contains messages with different original inputs.
This also allows the default batch size of the event source mapping for the continuing queue to be set to 10

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/handler.py
+++ b/handlers/aws/handler.py
@@ -218,8 +218,6 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
             processing_sent_event += 1
 
             if lambda_context is not None and lambda_context.get_remaining_time_in_millis() < _completion_grace_period:
-                processing_composing_shipper.flush()
-
                 return True, processing_sent_event
 
             return False, processing_sent_event
@@ -288,7 +286,7 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
                 continue
 
             if event_input.id in composite_shipper_cache:
-                composite_shipper = composite_shipper_cache[event_input.id]
+                composite_shipper = composite_shipper_cache[input_id]
             else:
                 composite_shipper = get_shipper_from_input(
                     event_input=event_input,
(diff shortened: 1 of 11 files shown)


handlers/aws/handler.py
```diff
@@ -202,87 +204,148 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
...
-
-            composite_shipper.send(es_event)
-            sent_event += 1
+    if trigger_type == "s3-sqs" or trigger_type == "sqs":
+        composite_shipper_cache: dict[str, CompositeShipper] = {}
 
-            if current_sqs_record > previous_sqs_record:
-                sqs_record = lambda_event["Records"][previous_sqs_record]
-                delete_sqs_record(trigger_event_source_arn, sqs_record["receiptHandle"])
+        def event_processing_callback(
```
zmoog, line 210: Pretty minor: I find the term "callback" in the name `event_processing_callback()` a little distracting, my mind keeps asking "where is it used as callback function?".

zmoog (approved):
Thanks for adding me as a reviewer; it was pretty interesting going through the whole project.

As a general comment, non-related to this specific PR, I found the function `lambda_handler()` pretty big, but I can also understand why.

I don't have any actionable advice right now, but please keep me involved in this project!

handlers/aws/handler.py
aspacca: you are right, I will remove the suffix

aspacca:
> As a general comment, non-related to this specific PR, I found the function `lambda_handler()` pretty big, but I can also understand why.

Indeed it is. It was much smaller when the project started

it could be split with no much effort but a lot of local variable visibility would be lost, if compensating this loss moving the local variable to params doesn't impact on readability it should be refactored for sure

PR #97, opened 2022-03-21
Include exlude support

### LINKED ISSUE #86: Add functionality of conditional include and exclude of events, based on regular expression based matching
Currently elastic-serverless-forwarder includes all events that are available in a given source input.
This issue is to add regular expression based include and/or exclude functionality where the logs are excluded or included based on matching the regular expression.
- Exclude is applied first and then include is applied, so exclude takes precedence if both are supplied by the user
- Filtering is applied to the original message
- User can specify one or more filters
- Debug logs should indicate what is being filtered by the forwarder
- The summary message should be present in the log by default providing how many events were skipped by the forwarder

aspacca:
Enhancement

## What does this PR do?
Add include/exclude feature

## Why is it important?
Users may have the need to skip some content to ingest, providing them include and exclude rules based configuration will let them achieve the gaol

## Checklist

## Author's Checklist

- [ ]

## Related issues

-
Closes #86

### DIFF
--- a/share/include_exlude.py
+++ b/share/include_exlude.py
@@ -1,0 +1,126 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+from __future__ import annotations
+
+import re
+from typing import Any, Optional
+
+
+def _get_by_key_path(key_path: str, data: dict[str, Any]) -> Any:
+    path_list = key_path.split(".")
+    result: Any = data
+    for key in path_list:
+        try:
+            coerced_key = int(key) if key.isnumeric() else key
+            result = result[coerced_key]
+        except KeyError:
+            result = None
+            break
+
+    return result
+
+
+class IncludeExcludeRule:
+    """
+    IncludeExcludeRule represents a pattern rule for a path key
+    """
+
+    def __init__(self, path_key: str, pattern: str):
+        self.path_key = path_key
+        self.pattern = re.compile(pattern)
+
+    def __eq__(self, other: object) -> bool:
+        assert isinstance(other, IncludeExcludeRule)
+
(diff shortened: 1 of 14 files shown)


endorama (approved):

share/include_exlude.py
```diff
@@ -0,0 +1,105 @@
...
+    def include_rules(self, value: list[IncludeExcludeRule]) -> None:
+        self._include_rules = value
+
+    @property
+    def exclude_rules(self) -> Optional[list[IncludeExcludeRule]]:
+        return self._exclude_rules
+
+    @exclude_rules.setter
+    def exclude_rules(self, value: list[IncludeExcludeRule]) -> None:
+        self._exclude_rules = value
```
zmoog, line 105: @aspacca the two `include_rules` and `exclude_rules` setters do not update the `_include_only`, `_exclude_only` and `_always_yield` values, this might lead to unexpected behaviour in the `filter()` method (probably an edge case):

```python
In [78]: f=IncludeExcludeFilter(include_patterns=[IncludeExcludeRule(pattern="a"), IncludeExcludeRule(pattern="b")])

In [79]: f.filter("a")
Out[79]: True

In [80]: f.filter("b")
Out[80]: True

In [81]: f.exclude_rules = [IncludeExcludeRule(pattern="b")]

In [82]: f.filter("a")
Out[82]: True

In [83]: f.filter("b")
Out[83]: True  # False expected here
```

I see the current workflow does not suggests this scenario, and setters are not used outside the constructor right now, so it's just a potential issue.

zmoog: I'm sorry for being late to this review party 😞

aspacca: yes, the `@property` in the object are just for accessing them in the tests for assertion

maybe we could raise on the setters to explicitly deny this possibility, what do you think?

we don't want to allow the setter as interface for mutating the rules

zmoog: Yep, LGTM!

zmoog (commented):
It's and edge case, we can handle this with no hurry.

PR #104, opened 2022-04-12
Fix continuing queue handle timeout

aspacca:
Bug

## What does this PR do?
Fix bug about improper handling of a message in the continuing queue originated from the continuing queue itself

## Why is it important?
Continuing queue was tested and handled only when the origin of the messages where one of the direct triggers.
Without handling the scenario where the origin messages are from the continuing queue itself the lambda does not behave properly

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/handler.py
+++ b/handlers/aws/handler.py
@@ -4,7 +4,7 @@
 
 import json
 import os
-from typing import Any, Callable
+from typing import Any, Callable, Optional
 
 from aws_lambda_typing import context as context_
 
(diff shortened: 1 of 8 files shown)


handlers/aws/handler.py
```diff
@@ -304,21 +316,24 @@ def handle_timeout(
 
             is_continuation_of_cloudwatch_logs = is_continuing_of_cloudwatch_logs(sqs_record)
 
+            input_id = sqs_record["eventSourceARN"]
```
kaiyan-sheng, line 319: just want to make sure we dont need to check if "eventSourceARN" in sqs_record here?

kaiyan-sheng (approved):

handlers/aws/handler.py
aspacca: `eventSourceARN` is always present in the record: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

PR #106, opened 2022-04-15
move autodiscovery at handler level

aspacca:
Enhancement
## What does this PR do?
Move the integration autodiscovery feature out of the shipper component to the input component: it is now the handler responsible to set an autodiscovery callable to the input

## Why is it important?
Moving the autodiscovery feature at the input level make the visibility of the integration scope available early in the code, allowing us to unlock feature like extracting records fields for cloudtrail in the storage, since now it will be possible to know at that point that the content is cloudtrail one

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -122,10 +122,54 @@ def wrap_try_except(
     return wrapper
 
 
+def discover_integration_scope(lambda_event: dict[str, Any], at_record: int) -> str:
+    if (
+        "Records" not in lambda_event
+        or len(lambda_event["Records"]) < at_record
+        or "body" not in lambda_event["Records"][at_record]
+    ):
+        return "generic"
+
+    body: str = lambda_event["Records"][at_record]["body"]
+    s3_object_key: str = ""
+    try:
+        json_body: dict[str, Any] = json.loads(body)
+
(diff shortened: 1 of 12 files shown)


handlers/aws/utils.py
```diff
@@ -122,10 +122,54 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+            return "aws.cloudwatch_logs"
+        elif "/elasticloadbalancing/" in s3_object_key:
+            return "aws.elb_logs"
+        elif "/network-firewall/" in s3_object_key:
+            return "aws.firewall_logs"
+        elif "/vpcflowlogs/" in s3_object_key:
+            return "aws.vpcflow"
+        elif "/WAFLogs/" in s3_object_key:
+            return "aws.waf"
+        else:
```
kaiyan-sheng, line 164: I wonder if we should add a log /warning message here to tell user here are the 6 log type we support and the rest goes into generic.

kaiyan-sheng (approved):

handlers/aws/utils.py
aspacca: we could have but not here in my opinion. rather in the elasticsearch sipper: I will do

PR #110, opened 2022-04-25
Support for cloudtrail logs

### LINKED ISSUE #102: SAR: support cloudtrail logs
**Describe the enhancement:**
Cloudtrail log entries are a single json object that we should extract the content of `Records` field from: it contains an array of json object that we should forward as events instead of the encompassing json log entry object

**Describe a specific use case for the enhancement or feature:**
Sending the json objects inside the array contained in `Records` field of the json log entry object of cloudtrail logs will allow the users to use the `aws.cloudtrail` integration

aspacca:
Enhancement
## What does this PR do?
Add support for extracting events from a json fields

## Why is it important?
Cloudtrail log entries are a single json object that we should extract the content of `Records` field from: it contains an array of json object that we should forward as events instead of the encompassing json log entry object

## Checklist

## Author's Checklist

- [ ]

## Related issues

-
Closes #102

### DIFF
--- a/handlers/aws/cloudwatch_logs_trigger.py
+++ b/handlers/aws/cloudwatch_logs_trigger.py
@@ -9,10 +9,11 @@ from typing import Any, Iterator
 
 from botocore.client import BaseClient as BotoBaseClient
 
-from share import shared_logger
+from share import extract_events_from_field, shared_logger
 from storage import CommonStorage, StorageFactory
 
 from .event import _default_event
+from .utils import extractor_events_from_field
 
 
 def _from_awslogs_data_to_event(awslogs_data: str) -> Any:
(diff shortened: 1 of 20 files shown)


handlers/aws/cloudwatch_logs_trigger.py
```diff
@@ -94,25 +96,28 @@ def _handle_cloudwatch_logs_event(event: dict[str, Any], aws_region: str) -> Ite
...
             assert isinstance(log_event, bytes)
 
-            es_event = deepcopy(_default_event)
-            es_event["@timestamp"] = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
-            es_event["fields"]["message"] = log_event.decode("UTF-8")
+            for extracted_log_event, extracted_starting_offset, is_last_event_extracted in extract_events_from_field(
+                log_event, json_object, starting_offset, ending_offset, integration_scope, extractor_events_from_field
+            ):
+                es_event = deepcopy(_default_event)
+                es_event["@timestamp"] = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
zmoog, line 106: I am not 100% sure about this, it's just a double check.

The use of the `Z` in the  `%Y-%m-%dT%H:%M:%S.%fZ` format implies this is a UTC timezone, right?

The use of `datetime.datetime.now()` to get a `datetime` object might return a local date and time.

aspacca: good catch!

it is not only here, btw, but on every input handler

I've chosen for format mimicking filebeat, forgetting about the timezone implications

we should use `datetime.datetime.utcnow()`

zmoog: I was bitten from Python's timezone support in the past ^^

zmoog (approved):

PR #123, opened 2022-06-15
Warn on shipper errors

aspacca:
Enhancement

## What does this PR do?
It logs as warning ingestion failures and fix minor formatting issues

## Why is it important?
When an event fails to be ingested in ES, it goes to the replay queue. The only way so far for the users to find about ingestion failure was to check if the replay queue was populated. This will allow them to know the specific events that failed to be ingested but will loose information about the reason of the failure. Adding a warning log when the failures happen with the reason of the failure as returned by ES will improve the discoverability of ingestion problems for the users.

## Checklist

~- [ ] I have added an entry in `CHANGELOG.md`~

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ docker-integration-test:  ## Run integration tests on docker
 docker-integration-test: SCRIPTS_BASE_DIR=docker/
 docker-integration-test: integration-test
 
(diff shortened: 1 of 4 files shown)


zmoog (approved):
Helpful for troubleshooting. LGTM.

PR #128, opened 2022-06-20
Handle properly gzip content regardless of the content type in S3 sto…

aspacca:
…rage

Bug
## What does this PR do?
Fix a false positive that leads to try to parse a gzipped plaintext content 

## Why is it important?
We don't rely on content-type for discovering gzip content rather instead on the header of the file, but still we use the content-type value for driving the strategy about inflating gzip content. there can be the case where a file is marked as `content-type: application/x-gzip` while it is not true. this leads to an error since we try to inflate the content.
this PR removes content-type out of the implementation.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/storage/payload.py
+++ b/storage/payload.py
@@ -27,7 +27,7 @@ class PayloadStorage(CommonStorage):
     @by_lines
     @inflate
     def _generate(
-        self, range_start: int, body: BytesIO, content_type: str, content_length: int
+        self, range_start: int, body: BytesIO, is_gzipped: bool, content_length: int
     ) -> Iterator[tuple[Union[StorageReader, bytes], Optional[dict[str, Any]], int, int, int]]:
         """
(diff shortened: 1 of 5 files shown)


storage/payload.py
```diff
@@ -54,19 +54,18 @@ def get_by_lines(
...
 
-        content_type = "plain/text"
-
         try:
             base64_decoded = base64.b64decode(self._payload, validate=True)
         except binascii.Error:
             base64_decoded = self._payload.encode("utf-8")
 
+        is_gzipped: bool = False
         if base64_decoded.startswith(b"\037\213"):  # gzip compression method
```
zmoog, line 63: ~Friends~ Bytes don't lie.

zmoog (approved):

PR #145, opened 2022-08-03
Refactor json collection and event expander

### LINKED ISSUE #141: Let the user provide an hint on the format of json content
**Describe the enhancement:**
Introduce a `json_content_type` setting at input level where the user can provide an hint on the format of the json content: either a single json object (spanning or not multiple lines) per payload, or `ndjson` format per payload

**Describe a specific use case for the enhancement or feature:**

The Elastic Serverless Forwarder is able to discovery automatically JSON content in the payload of an input and collect the contained JSON objects in the payload.
The JSON objects can be either be on a single line or spanning multiple lines. In the second case the Elastic Serverless Forwarder expects the different JSON objects spanning multiple lines to be separated by a newline.

In case of JSON objects spanning multiple lines a limit of 1000 lines is applied: every JSON object spanning on more lines than those will not be collected. Every line composing the whole JSON object will be forwarder individually instead,

Sometimes relaying on the Elastic Serverless Forwarder JSON content auto-discovery feature might have a huge impact on performance, or you have a known payload content of a single JSON object spanning more than 1000 lines. In this case you can provide in the input configuration and hint on the nature of the JSON content: this will change the parsing logic applied and improve performance or overcome the 1000 lines limit.

aspacca:
Enhancement
## What does this PR do?
see #141 

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

Closes #141

### DIFF
--- a/docs/README-AWS.md
+++ b/docs/README-AWS.md
@@ -463,18 +463,18 @@ No flags are used when the regular expression is compiled. Please refer to [inli
 
 ## JSON content discovery
 
-The Elastic Serverless Forwarder is able to discovery automatically JSON content in the payload of an input and collect the contained JSON objects in the payload.
-The JSON objects can be either be on a single line or spanning multiple lines. In the second case the Elastic Serverless Forwarder expects the different JSON objects spanning multiple lines to be separated by a newline.
+The Elastic Serverless Forwarder is able to discovery automatically JSON content in the payload of an input and collect the JSON objects contained in the payload.
+The JSON objects can be either be on a single line or spanning multiple lines. In the second case the Elastic Serverless Forwarder expects each different JSON objects spanning multiple lines to be separated by a newline delimiter.
 
-In case of JSON objects spanning multiple lines a limit of 1000 lines is applied: every JSON object spanning on more lines than those will not be collected. Every line composing the whole JSON object will be forwarder individually instead,
+In case of JSON objects spanning multiple lines a limit of 1000 lines is applied: every JSON object spanning across more 1000 lines will not be collected. Every line composing the whole JSON object will be forwarded individually instead.
 
 Sometimes relaying on the Elastic Serverless Forwarder JSON content auto-discovery feature might have a huge impact on performance, or you have a known payload content of a single JSON object spanning more than 1000 lines. In this case you can provide in the input configuration and hint on the nature of the JSON content: this will change the parsing logic applied and improve performance or overcome the 1000 lines limit.
 
 **Notes:**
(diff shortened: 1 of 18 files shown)


endorama (approved):

endorama:
I'm wondering if the auto discovery mechanism may be deprecated, to reduce our code footprint. If we allow the use to specify the content of the payload, couldn't we choose a sane default (is there one?) and let the user configure it for their specific usage when needed?

aspacca:
> I'm wondering if the auto discovery mechanism may be deprecated, to reduce our code footprint. If we allow the use to specify the content of the payload, couldn't we choose a sane default (is there one?) and let the user configure it for their specific usage when needed?

To deprecate auto discovery mechanism in order to reduce the code footprint we cannot apply any default: the user will have to tell us at least that the content is of json type.
That would not be enough to remove all the code related to parsing a growing buffer in multiple iterations: we'd need to know upfront if the json format is a single json object or an `ndjson` (the "hint" provided by the `json_content_type` setting introduced here, that in case will become mandatory )

Only in this case we could simplify a lot the code, but this will left out the case of multiple json objects, each of them spanning multiple lines, separated by a newline delimiter.
It could be the case to drop supporting such format, if we have enough evidence it's not a meaningful use case to support.

At least with this PR we have a way to improve performance for the other two cases. :)

endorama:
> At least with this PR we have a way to improve performance for the other two cases. :)

I agree! But it feels weird that we default to the less performant implementation, requiring user action to increase performances. If the most common case is either single or ndjson we may want to increase performance for that case instead! Would be nice having some metrics about this hint field value when deployed by customer ;)

aspacca:
> But it feels weird that we default to the less performant implementation

It will be the less performant in the case of single json spanning multiple lines.
For `ndjson` and the other remaining format it is as much performant as before (while the code footprint can be reduced for `ndjson` with the micro-optimisation this will carry)

We default to this because it's the only implementation that supports all the three formats (or all the two, if we decide to remove support for the last use case)

To summarise I see two different topics for improvement here:
1. reduce the complexity of the code
2. always run the most performant implementation for the specific json format

The first is a consequence of the second: as that for achieving the second goal we must ask the user to provide the json format the forwarder will parse. Having this information will reduce code complexity as well, to what extent it depends on support 2/3 or 3/3 of the possible formats :)

aspacca:
> Would be nice having some metrics about this hint field value when deployed by customer ;)

coupled with the metrics about the actual formats distribution 👍

PR #148, opened 2022-08-03
Optimise ESF

aspacca:
Enhancement

## What does this PR do?

- Add benchmark for covering `expand_event_list_from_field` and json dumper
- Fix `json_content_type` not passed as argument to storage factories
- Set custom json parser and dumper in es client and for storage decorator and event expander
- Add `disabled` value for `json_content_type` in order to totally skip json auto discovery

## Why is it important?

We need to optimise as much as possible handling json content, since the most performance impacting code in the forwarder.
Having benchmarks available for the different use cases will help decide which json library to use and in case even switching to different libraries according to the matrix of `expand_event_list_from_field` and `json_content_type` setting provided by the user

I've run the new benchmark and identified ujson` as the most performing json package

Here the outcome of several iterations of optimisation:
```
320 MB, 10mins timeout, batch_max_actions: 4000, batch_max_bytes: 104857600
1.3.0-rc20
Sent Events: 618500, Duration: 480secs, Max Memory Used: 97MB
Sent Events: 445500, Duration: 341secs, Max Memory Used: 100MB

1.3.0-rc13
Sent Events: 580000, Duration: 481secs, Max Memory Used: 135MB
Sent Events: 484000, Duration: 404secs, Max Memory Used: 138MB

1.3.0-rc8
Sent Events: 364000, Duration: 482secs, Max Memory Used: 131MB
Sent Events: 364000, Duration: 481secs, Max Memory Used: 131MB
Sent Events: 336000, Duration: 445secs, Max Memory Used: 132MB

1.3.0-rc6
Sent Events: 356000, Duration: 480secs, Max Memory Used: 128MB
Sent Events: 356000, Duration: 483secs, Max Memory Used: 131MB
Sent Events: 352000, Duration: 477secs, Max Memory Used: 131MB

1.2.1
Sent Events: 340000, Duration: 483secs, Max Memory Used: 126MB
Sent Events: 340000, Duration: 483secs, Max Memory Used: 127MB
Sent Events: 340000, Duration: 483secs, Max Memory Used: 127MB
Sent Events:  44000, Duration: 285secs, Max Memory Used: 127MB

```

## Checklist

## Author's Checklist

- [ ]

## Related issues

-


storage/payload.py
```diff
@@ -62,20 +62,33 @@ def chunk_lambda() -> Any:
...
 
+        is_gzipped: bool = False
+        is_b64encoded: bool = False
         try:
             base64_decoded = base64.b64decode(self._payload, validate=True)
             if not base64_decoded.startswith(b"\037\213"):  # gzip compression method
                 base64_decoded.decode("utf-8")
+                is_b64encoded = True
+            else:
+                is_b64encoded = True
```
tommyers-elastic, line 73: should this be `False`?

storage/payload.py
```diff
@@ -62,20 +62,33 @@ def chunk_lambda() -> Any:
     def get_by_lines(self, range_start: int) -> Iterator[tuple[bytes, int, int, Optional[int]]]:
         original_range_start: int = range_start
 
+        is_gzipped: bool = False
+        is_b64encoded: bool = False
         try:
             base64_decoded = base64.b64decode(self._payload, validate=True)
             if not base64_decoded.startswith(b"\037\213"):  # gzip compression method
```
tommyers-elastic, line 69: nit: and not related to the PR - would be good to remove duplication of this check (L77, L111)

storage/storage.py
```diff
@@ -8,7 +8,7 @@
 
 from share import ExpandEventListFromField, ProtocolMultiline
 
-CHUNK_SIZE: int = 1024
+CHUNK_SIZE: int = 1024**2
```
tommyers-elastic, line 11: this is quite a significant change - a comment could be useful as to the significance of this number

tommyers-elastic (approved):
nice work ! couple of comments, no blockers.

handlers/aws/handler.py
```diff
@@ -213,10 +210,12 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
...
             current_kinesis_record_n,
         ) in _handle_kinesis_record(
-            lambda_event, event_input.id, expand_event_list_from_field, event_input.get_multiline_processor()
+            lambda_event,
+            event_input.id,
+            expand_event_list_from_field,
+            event_input.json_content_type,
+            event_input.get_multiline_processor(),
         ):
-            shared_logger.debug("es_event", extra={"es_event": es_event})
```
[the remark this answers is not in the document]
aspacca: they expose sensitive data

storage/payload.py
aspacca: good catch!

storage/storage.py
aspacca: yes indeed!

storage/payload.py
aspacca: I'm not sure we can remove duplicating the check
we have the following scenarios:
1- valid base64encode of gzip content
2- valid base64encode of plaintext content
3- bare gzip content
4- plaintext that will not raise a `binascii.Error` while base64 decoding but still does not produce valid unicode content.

4 it's tricky, and I've found totally by chance by sending some plaintext content (I don't remember exactly, whatever like "text blah blah"). this was base 64 encode to not unicode string, failing later on when try to decode in the decorators

that's why I added [L70](https://github.com/elastic/elastic-serverless-forwarder/pull/148/files/ad13910c8fc50a5865d8bd891806b48d6eae750b#diff-11a8f6bb928926e1b10ead6bb98a7adbea55f0590cb44c18d7094e7f0d883b15R70), but I have to skip if it's gzip content, because that will raise the `UnicodeDecodeError`

share/json.py
```diff
@@ -0,0 +1,15 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+from typing import Any, AnyStr
+
+import ujson
+
+
+def json_dumper(json_object: Any) -> str:
```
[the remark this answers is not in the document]
aspacca: that will be hard to maintain when we want to mock for benchmark etc, I suspect

also different libraries have different arguments

in the case we refactor like you propose will have to mock change the implementation of `WrappedJson`?
but this will require in the end writing N `WrappedJson*` like we have N `json_dumper*`, did I get it right?

what will make the difference is using `from shared import json` instead of `import json`, but this indeed for me it's not explicit enough about the fact that we are indeed using a custom json lib and it's error prone since I see we could easily forget to import json from `shared` instead that from the stdlib

since we are not providing a package to be consumed from third party developers, having a more explicit `json_dumper` make me more comfortable on this

not sure, we can discuss further and add the refactoring later on

handlers/aws/cloudwatch_logs_trigger.py
```diff
@@ -123,23 +122,30 @@ def _handle_cloudwatch_logs_event(
...
-
-            es_event["fields"]["log"]["file"]["path"] = f"{log_group_name}/{log_stream_name}"
-
-            es_event["fields"]["aws"] = {
-                "cloudwatch": {
-                    "log_group": log_group_name,
-                    "log_stream": log_stream_name,
-                    "event_id": event_id,
-                }
+            es_event: dict[str, Any] = {
```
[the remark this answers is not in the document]
aspacca: every `es_event` in every trigger is slight different (specifically the part related to the `aws` payload)
I will look into abstracting this without ending up on passing around a lot of specific arguments

in general I would like to abstract/clean the whole `handler.py` and the different triggers logic code

I will keep for a future refactoring

storage/payload.py
tommyers-elastic: was mainly just thinking it would be good to have `startswith(b"\037\213")` refactored into a function

storage/payload.py
tommyers-elastic: got it - in this case then it would be good to refactor this, it looks pretty weird `if foo: bar = True; else bar = False`...

share/json.py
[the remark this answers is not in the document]
tommyers-elastic: we have the issue of forgetting to use the shared json lib regardless of how we name/implement the wrappers. one mitigation for this is instrumenting the import of stdlib `json` at test time and throwing an exception. another is adding some additional comments into the shared json lib that explains the importance of ujson vs stdlib for proper functioning of the forwarder.

the mocking piece seems fairly trivial either way - in this case we could inject the base lib into the constructor, and for any different arguments just overwrite the affected members. 

all that said i really don't mind how we do it. my main thing was that we are introducing really non-standard naming. in particular the `json_dumper/loader` names reads to me like they're factories. but it's not worth spending too much time on. i'm happy with this as is, so we can leave it.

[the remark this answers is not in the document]
aspacca: > we have the issue of forgetting to use the shared json lib regardless of how we name/implement the wrappers. one mitigation for this is instrumenting the import of stdlib `json` at test time (or even runtime) and throwing an exception. another is adding some additional comments into the shared json lib that explains the importance of ujson vs stdlib for proper functioning of the forwarder.

what I mean is that looking through the code having `json_dumper`/`json_loader` it will be a more evident hint that we are not using the stdlib `json` or we are wrapping it at some degree.
also I have a current branch testing [hyperscan](https://github.com/intel/hyperscan) for the multiline and include/exclude filter, and while the API is different from stdlib `re`, even if they would be the same, there is a case for the AWS secret manager expander, where I don't want to replace them. this lead me think that similarly there could be some case where we will be fine with stdlib `json` and in this case it would just be a too much subtle for me `import json` vs `from share import json` to make the difference :)

> all that said i really don't mind how we do it. my main thing was that we are introducing really non-standard naming. in particular the `json_dumper/loader` names reads to me like they're factories. but it's not worth spending too much time on. i'm happy with this as is, so we can leave it.

as you can see my main concern regarding the comment was about readability :)
In this regard the fact that we use non-standard naming is somehow intended, as said to hint about using non-standard module.
at the same time if `json_dumper/loader` names read like they're factories, I'm totally open to find better names :)

No need to block merging for this, unless you have a prompt alternatives :), but let's improve the naming in the future

storage/payload.py
aspacca: I will add comments to explain the behaviour for the moment and think about a way to refactor in the future

aspacca:
fixes #151

Issue #50, opened 2021-12-15
SAR: Add telemetry data for SAR usage


ravikesarwani:
Add telemetry data for SAR usage. Initial focus is on usage of the Lambda function itself and the inputs being used. Things like:
- Some unique identifier of the deployed Lambda from SAR
- What input is being used (SQS S3, Kinesis, CloudWatch)
We need to collect and send data only once per SAR execution (Lambda function be default usually runs for 15 minutes).

ruflin:
Here is a related issue: https://github.com/elastic/obs-dc-team/issues/722 The challenge is how we collect and ship the data. For the lambda use case, I think we would also need to ship to the telemetry cluster directly. Pulling in @afharo here as he chimed in on the other thread too.

damien:
As a user of ESF, would this telemetry be collected through Elastic Cloud or are you folks looking to send telemetry directly from the Lambda function?

If it's the latter I'd suggest making this opt-out or ensuring that it fails safely. My organization, and likely others, keeps a close eye on egress traffic from our cloud infrastructure. This means any lambda making connections to new endpoints would trigger an alert.

Furthermore, a lot of deployments have whitelisted endpoints that traffic is allowed to flow to. Instrumentation connecting to unknown hosts will be blocked in many contexts, which if not accommodated for will raise errors for ESF deployments.

afharo:
@ruflin thank you for the ping. To be fair, I'm surprised AWS does not provide any stats around how many "installations" there are for a public Lambda. IMO, knowing how popular an application is would serve both the developer and the users. I Amazon provided that feature, it would cover most of the insights we are after.

Given that AWS does not help here, we need to design our own solution to this. 

## Premises

### Minimal impact on costs to the users

I agree with @damien that we should look for a solution that does not incur additional costs to the users (or the impact is minimal). That's a must in all telemetry we collect.

### Minimize external requests

It's highly possible that the `elastic-server-forwarder` is connected to a private VPC. It implies that the users will need to explicitly allow it to reach the Internet ([guide](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/)). The same would apply if they maintain an allow-list, as also mentioned in the previous comment.

We should aim to collect this information without making any additional requests to other URLs.

## Proposal

### How to "ship" the information?

[...]

PR #156, opened 2022-08-26
Telemetry

aspacca:
## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates #50

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -18,7 +18,6 @@ from share import (
     Input,
     Output,
     get_hex_prefix,
-    function_ended_telemetry,
     input_has_output_type_telemetry,
     json_dumper,
     json_parser,
@@ -107,7 +106,7 @@ def wrap_try_except(
 
             shared_logger.exception("exception raised", exc_info=e)
 
-            function_ended_telemetry(exception_raised=True)
+            # function_ended_telemetry(exception_raised=True)
 
             raise e
 
@@ -120,7 +119,7 @@ def wrap_try_except(
 
             shared_logger.exception("exception raised", exc_info=e)
 
-            function_ended_telemetry(exception_ignored=True)
+            # function_ended_telemetry(exception_ignored=True)
 
             return f"exception raised: {e.__repr__()}"
 
@@ -157,6 +156,7 @@ def get_shipper_from_input(event_input: Input, config_yaml: str) -> CompositeShi
         anonymized_arn = anonymize_arn(event_input.id)
         input_has_output_type_telemetry(
             input_id=anonymized_arn.id,
+            input_type=event_input.type,
             output_type=output_type,
         )
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -521,36 +521,36 @@ def expand_event_list_from_field_resolver(integration_scope: str, field_to_expan
 
 
 @dataclass
-class ARN(object):
+class ARN:
     """The Amazon Resource Name (ARN) of an AWS resource."""
 
-    id: str
-    service: str
-    region: str
-    account_id: str
+    def __init__(self, arn: str) -> None:
+        arn_components = arn.split(":")
+        self.arn = arn
+        self.service = arn_components[2]
+        self.region = arn_components[3]
+        self.account_id = arn_components[4]
+        self.resource_id = arn_components[5]
 
+    @property
+    def hashed_arn(self) -> str:
+        return hashlib.sha256(self.arn.encode("utf-8")).hexdigest()[:10]
 
-def anonymize_arn(arn: str) -> ARN:
-    """Anonymize an ARN by hashing the account ID and the resource ID."""
-    arn_components = arn.split(":")
-
-    hashed_arn = hashlib.sha256(arn.encode("utf-8")).hexdigest()[:10]
-    service = arn_components[2]
-    region = arn_components[3]
-    account_id = hashlib.sha256(arn_components[4].encode("utf-8")).hexdigest()[:10]
-
-    return ARN(hashed_arn, service, region, account_id)
+    @property
+    def hashed_resource_id(self) -> str:
+        return hashlib.sha256(self.resource_id.encode("utf-8")).hexdigest()[:10]
 
 
 def build_function_context(lambda_context: context_.Context) -> FunctionContext:
     """Create a FunctionContext from a Lambda context."""
-    anonymized_arn = anonymize_arn(lambda_context.invoked_function_arn)
+
+    arn = ARN(lambda_context.invoked_function_arn)
 
     return FunctionContext(
-        anonymized_arn.id,
+        arn.hashed_arn,
         lambda_context.function_version,
         lambda_context.aws_request_id,
-        anonymized_arn.region,
+        arn.region,
         "aws",
         lambda_context.memory_limit_in_mb,
     )
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -12,13 +12,12 @@ from botocore.client import BaseClient as BotoBaseClient
 from elasticapm import Client
 from elasticapm import get_client as get_apm_client
 from elasticapm.contrib.serverless.aws import capture_serverless as apm_capture_serverless  # noqa: F401
-from share import (  # function_ended_telemetry,
+from share import (  # function_ended_telemetry,; input_has_output_type_telemetry,
     Config,
     FunctionContext,
     Input,
     Output,
     get_hex_prefix,
-    input_has_output_type_telemetry,
     json_dumper,
     json_parser,
     shared_logger,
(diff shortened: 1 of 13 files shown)


zmoog:
Hey, @aspacca, I just pushed a commit with changes to the number of events tracked by the telemetry module.

The idea is to follow these principles:

- send only a core set of data (function id and input source)
- avoid delaying the lambda execution, if possible.

With this approach, the telemetry worker can send the data after successfully parsing the configuration while the main thread is working on the business data.

Waiting longer to get additional telemetry data like events stats and execution time is optional and adds seconds to the general lambda execution time. We can still add it later if we need to.

cc @tommyers-elastic

zmoog:
I would split the current development in:

- phase 1: ship the telemetry feature with an essential data set (function id and event source) and validate the approach.
- phase 2: add additional telemetry data, if needed. 

@aspacca @tommyers-elastic, let me know what you think.

handlers/aws/utils.py
```diff
@@ -174,6 +174,7 @@ def get_shipper_from_input(
         anonymized_arn = anonymize_arn(event_input.id)
         input_has_output_type_telemetry(
             input_id=anonymized_arn.id,
+            input_type=event_input.type,
```
aspacca, line 177: `get_shipper_from_input` is called event by event for a single forwarder execution

we have three cases:
1. the original trigger is either `kinesis-data-stream`, `s3-sqs` or `sqs` (but not the continuing queue). in this case all the calls are idempotent
2. the original trigger is a `cloudwatch-logs`. in this case if the ARN of the input is that of a log group, all the calls are idempotent. at the contrary if the ARN is that of a log stream, the calls are not idempotent
3. the original trigger is the specific case of the `sqs` continuing queue. in this case the `event_input` object is indeed the one referencing the continued original trigger. in the same execution it can be any (`cloudwatch-logs`, `kinesis-data-stream`, `s3-sqs` or `sqs`). the calls are not idempotent

there are two things to consider here:
at https://github.com/elastic/elastic-serverless-forwarder/pull/156/commits/ee0b5a88bf8a63b4fd706121b3156d4c5059fcb4#diff-6fde71d82d3ead64aa933407ae918d5d91fbb4be69e09f236c2d5579c2cdb733R349 we send telemetry data every time `input_has_output_type_telemetry` is invoked: this happens multiple times. with that instead of sending a single telemetry call we send has much as the size of the lambda event.
moreover every single telemetry call will have the same payload where the data relative to the output gets growing, according if we are in one of the non idempotent cases

> the telemetry worker can send the data after successfully parsing the configuration while the main thread is working on the business data.

I'm fine with this approach, but then we should have
```
if isinstance(event, FunctionStartedEvent):
```

at https://github.com/elastic/elastic-serverless-forwarder/pull/156/commits/ee0b5a88bf8a63b4fd706121b3156d4c5059fcb4#diff-6fde71d82d3ead64aa933407ae918d5d91fbb4be69e09f236c2d5579c2cdb733R349

changing to `InputHasOutputTypeEvent` drastically increases the number of times we send data to the telemetry server :)

share/telemetry.py
```diff
@@ -373,54 +388,59 @@ def function_started_telemetry(ctx: FunctionContext) -> None:
...
-        with_exception = WithExceptionTelemetryEnum.EXCEPTION_IGNORED
-    elif exception_raised:
-        with_exception = WithExceptionTelemetryEnum.EXCEPTION_RAISED
-
-    telemetry_command = FunctionEndedEvent(with_exception=with_exception, to_be_continued=to_be_continued)
-    telemetry_queue.put(telemetry_command)
-
-
-def input_has_output_type_telemetry(input_id: str, output_type: str) -> None:
+# def function_ended_telemetry(
```
aspacca, line 391: minor: not sure if we need to introduce all this comments in order to not have the code running

for the final outcome of sending data to the telemetry only once at, just after parsing the config file, we can keep the code and change just https://github.com/elastic/elastic-serverless-forwarder/pull/156/commits/ee0b5a88bf8a63b4fd706121b3156d4c5059fcb4#diff-6fde71d82d3ead64aa933407ae918d5d91fbb4be69e09f236c2d5579c2cdb733R349

share/telemetry.py
```diff
@@ -85,18 +85,20 @@ class TelemetryData:
...
     output_sent_to_replay: dict[str, int] = {}
 
-    def add_output_type_for_input(self, input_id: str, output_type: str) -> None:
+    def set_output_type_for_input(self, input_id: str, input_type: str, output_type: str) -> None:
         """Add the output type for the input."""
         if input_id not in self.input_outputs_type:
-            self.input_outputs_type[input_id] = list()
-
-        self.input_outputs_type[input_id].append(output_type)
+            self.input_outputs_type[input_id] = {
```
aspacca, line 97: while we have only one "elasticsearch" output at the moment, we are going soon to add also the logstash one.
potentially every input can have as many outputs as the number of types we support.
this change seems to me that reduce to a single one the data related to the output for a given input

handlers/aws/utils.py
zmoog: Yeah, noticed the for loop a few lines above right after I pushed the changes ^^

I was considering sending the telemetry data after the config is successfully parsed.

share/telemetry.py
zmoog: I commented out a bunch of code to highlight the impact area. It can be handy to keep it around until we complete the review iteration.

I am okay with keeping the code around (uncommented) since this is primarily a telemetry phase 1, and we may go back to it later. However, usually, I remove unused code, even when it is inconvenient going back to previous PRs.

share/telemetry.py
zmoog: Oh, I forgot the Logstash output is coming 🤦  Thank you for the heads up!

share/telemetry.py
aspacca: > However, usually, I remove unused code

that's my preferred option as well :)

aspacca:
> The function triggers the `ConfigLoadedEvent` event right after the configuration is successfully fetched and parsed from its source.

I've only one minor concern: if an exception is raised when parsing the config (that's "legit") we lose sending any telemetry data

zmoog:
@aspacca After [working on the Indexer](https://github.com/elastic/telemetry/pull/2056/files), I realized the current payload might contain redundant data. 

Instead of sending the list of configured inputs, a better option is to send only the input sourcing the data for each lambda execution.

For example:

```JSON
{
  "function_id": "80036aa038",
  "function_version": "v0.0.0",
  "execution_id": "6b76b62589",
  "cloud_provider": "aws",
  "cloud_region": "us-east-1",
  "memory_limit_in_mb": "512",
  "start_time": "1676016439.435081",
  "input": {
    "type": "s3-sqs",
    "outputs": [
      "elasticsearch"
    ]
  }
}
```

Now the payload is focused on the lambda execution, but we can still get the idea of which inputs each deployment uses by putting together data from different lambda executions.

handlers/aws/utils.py
```diff
@@ -510,3 +529,39 @@ def expand_event_list_from_field_resolver(integration_scope: str, field_to_expan
...
+    service: str
+    region: str
+    account_id: str
+
+
+def anonymize_arn(arn: str) -> ARN:
+    """Anonymize an ARN by hashing the account ID and the resource ID."""
+    arn_components = arn.split(":")
+
+    hashed_arn = hashlib.sha256(arn.encode("utf-8")).hexdigest()[:10]
```
aspacca, line 548: hello @zmoog , beware that we experienced collision when truncating the sha256 hash to 10 chars, better use the full value every where we apply the hash

zmoog:
A couple of days ago, I received some feedback about the data model
from @aspacca.

After a brief discussion, we realized the current data model is only
adequate for some use cases.

Here's the current one:

    {
        "function_id": "80036aa038",
        "function_version": "v0.0.0",
        "execution_id": "6b76b62589",
        "cloud_provider": "aws",
        "cloud_region": "us-east-1",
        "memory_limit_in_mb": "512",
        "input": {
            "type": "s3-sqs",
            "outputs": [
                "elasticsearch"
            ]
        },
    }

As Andrea highlighted, there are circumstances (replay and continuing
queues) where a single lambda execution can involve multiple input
types. We need to have more than just a single input.

So I changed the data model to drop the `input` and `outputs` fields
and add a `inputs` list with all inputs involved in the execution.

And here's the updated data model:

    {
        "function_id": "80036aa038",
        "function_version": "v0.0.0",
        "execution_id": "6b76b62589",  # optional?
        "cloud_provider": "aws",
        "cloud_region": "us-east-1",
        "memory_limit_in_mb": "512",
        "inputs": [
            {"type": "s3-sqs", "output": "elasticsearch"},
            {"type": "cloudwatch-logs", "output": "elasticsearch"}
        ]
    }

zmoog:
Besides the data model, we also noticed that the telemetry function
`input_selected_telemetry` can be called multiple times when replay
and continuing queues are involved.

To avoid multiple calls to the Telemetry API, we added a new telemetry
event called `event_processed_telemetry` triggered when the lambda
function has processed all events and is about to flush the shippers
to Elasticsearch.

The telemetry sends the data to the telemetry endpoint on the
`event_processed_telemetry` event.

zmoog:
I added a new configuration parameter in the SAP application to enable or disable the telemetry.

The parameter is called `ElasticServerlessForwarderTelemetry`

It also adds a similar new configuration option for the script-based
install:

    telemetry:
        enabled: true
        endpoint: "https://telemetry.elastic.co/v3/send/esf"

handlers/aws/utils.py
```diff
@@ -94,6 +102,8 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
 
             shared_logger.exception("exception raised", exc_info=e)
 
+            # function_ended_telemetry(exception_raised=True)
```
aspacca, line 105: https://github.com/elastic/elastic-serverless-forwarder/blob/d5bce03e119b35f00e7f263bb6b38a3f43a70bdf/handlers/aws/utils.py#L88-L89
```
        try:
            ret = func(lambda_event, lambda_context)
            if ret == "continuing":
                    function_ended_telemetry(to_be_continued=True)
            else:
                    function_ended_telemetry(to_be_continued=False)

            return ret
```

zmoog:
We implemented telemetry by parsing the cloud proxy logs. Closing this PR.

PR #157, opened 2022-08-29
batch in shipper for replay handler

aspacca:
Enhancement
## What does this PR do?
Add support for batch send in shipper for replay queue handler

## Why is it important?
Currently the replay handler flush the send at every message processed by the replay queue.
this prevents the original batch settings of the replayed output to be applied, doing the contrary, like this PR take care of, improves performance

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/handler.py
+++ b/handlers/aws/handler.py
@@ -127,15 +127,23 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
 
     if trigger_type == "cloudwatch-logs":
         cloudwatch_logs_event = _from_awslogs_data_to_event(lambda_event["awslogs"]["data"])
-        log_group_arn, aws_region = get_log_group_arn_and_region_from_log_group_name(cloudwatch_logs_event["logGroup"])
-        input_id = log_group_arn
+        log_stream_arn, aws_region = get_log_group_arn_and_region_from_log_group_name(
+            cloudwatch_logs_event["logGroup"], cloudwatch_logs_event["logStream"]
+        )
+        input_id = log_stream_arn
 
         event_input = config.get_input_by_id(input_id)
 
         if event_input is None:
-            shared_logger.warning("no input defined", extra={"input_type": trigger_type, "input_id": input_id})
+            log_group_arn_components = log_stream_arn.split(":")
+            input_id = f"{':'.join(log_group_arn_components[:-2])}:*"
 
-            return "completed"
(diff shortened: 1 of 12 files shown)


handlers/aws/handler.py
```diff
@@ -81,22 +81,41 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
...
+                if shipper is None:
+                    shared_logger.warn(
+                        "no shipper for output in replay queue",
+                        extra={"output_type": event["output_type"], "event_input_id": event["event_input_id"]},
+                    )
+                    continue
+
+                shipper_cache[input_id] = shipper
+            else:
+                shipper = shipper_cache[input_id]
```
tommyers-elastic, line 106: under what circumstances would different events require different shippers?

aspacca: when the replay messages come from different inputs with different outputs

tommyers-elastic: got it - if we have a single lambda deployment configured, do we still end up constructing multiple shippers here, one per batch of input events, even though the configuration of the shipper is identical? 

i don't see any issues, just trying to understand if there's an optimization we could make (for example change the caching logic to be based on the shipper config, vs the input id).

aspacca: > i don't see any issues, just trying to understand if there's an optimization we could make (for example change the caching logic to be based on the shipper config, vs the input id).

yes, this is an optimisation that could be made, but most probably [here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/handlers/aws/handler.py#L372-L382), since in the replay queue the same shipper (as es cluster with credentials) can have different [es_datastream_name](https://github.com/elastic/elastic-serverless-forwarder/pull/157/files#diff-a50c2cea8bcb0bea90c7963fe6a2ecb9a44fe4a04cd8738ce7d9f626d2109ea6R54), according to the original input

aspacca: > yes, this is an optimisation that could be made, but most probably [here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/handlers/aws/handler.py#L372-L382)

even there the same related to `es_datastream_name` is true.
it would be relevant refactoring to try to optimise the shipper cache removing `es_datastream_name` from the picture, not sure if it is worth

tommyers-elastic (approved):
👍 all good - i still think there's room for optimization in terms of constructing fewer identical shippers, but happy for this to be addressed separately.

PR #166, opened 2022-10-03
Simplify the Shipper base class type 

tommyers-elastic:
The ProtocolShipperType was causing some confusion in our static type checking (mypy). Removing it and just using ProtocolShipper as the return type of the shipper factory simplifies this. Explicitly subclassing the ABC ('ProtocolShipper'), also guarantees a runtime check that the shipper class implements the required ABC methods.

The only downside of this approach is that we cannot write this code:

```
elasticsearch_shipper: ElasticsearchShipper = create_from_output('elasticsearch', ...)
```

This seems totally reasonable - sure, the factory returns different types depending on its inputs, but I think its better for callers not to assume which concrete type, and just use the ABC methods only.

### DIFF
--- a/handlers/aws/replay_trigger.py
+++ b/handlers/aws/replay_trigger.py
@@ -5,7 +5,7 @@
 from typing import Any, Optional
 
 from share import Config, ElasticsearchOutput, Input, Output, shared_logger
-from shippers import ElasticsearchShipper, ShipperFactory
+from shippers import ShipperFactory
 
 from .exceptions import InputConfigException, OutputConfigException, ReplayHandlerException
 from .utils import delete_sqs_record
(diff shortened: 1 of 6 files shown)


aspacca:
@tommyers-elastic I've already took care of it in both current open PRs where the linter is broken

but thank you for being on the spot :)

I just didn't subclassed from `ProtocolShipper` since it should not be the way to go as far as I understand.

tommyers-elastic:
thanks andrea!

> I just didn't subclassed from ProtocolShipper since it should not be the way to go as far as I understand.

subclassing an ABC provides runtime checks that a subclass implements all abstract methods. i don't see any reason not to do this even with the additional type checks. Otherwise it's not clear to the reader (at least, to me!) that ElasticsearchShipper implements ProtocolShipper.

aspacca:
> subclassing an ABC provides runtime checks that a subclass implements all abstract methods. i don't see any reason not to do this even with the additional type checks

because we are using `Protocol` and from what I get they should not be subclassed by concrete implementation https://peps.python.org/pep-0544:
> A concrete type X is a subtype of protocol P if and only if X implements all protocol members of P with compatible types. In other words, subtyping with respect to a protocol is always structural.

I indeed did a refactoring migrating from abstract classes to protocols :)

PR #183, opened 2022-11-16
[Logstash] Retry and Replay queue handling

girodav:
## What does this PR do?

This PR does the following:
- Adds Retry+Exponential backoff logic for the logstash shipper, leveraging `urllib3.util.retry`
- Adds basic logic to send failed events to the replay queue, if the maximum number of retries is exceeded. Since we cannot know which events failed to be processed by Logstash, the whole batch of events is sent to the replay queue.
- Adds "_id" to the event forwarded by the Logstash shipper, similarly to what done in the ElasticSearch shipper. 
    - _NOTE: This is not stricly required by Logstash, but was added to simplify the implementation, since the replay queue handler that is currently present, assumes that the event includes `_id`_
- Improves unit tests for the Logstash shipper, using `responses`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ elasticsearch>=7.16,<7.17
 PyYAML>=6.0,<6.1
 aws_lambda_typing>=2.16.1,<3
 ujson>=5.5.0,<5.6
-requests>=2.25.1
-
+requests<2.26,>=2.20.0
+responses>=0.22.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ elasticsearch>=7.16,<7.17
 PyYAML>=6.0,<6.1
 aws_lambda_typing>=2.16.1,<3
 ujson>=5.5.0,<5.6
-requests<2.26,>=2.20.0
+requests==2.25.1
 responses>=0.22.0
--- a/shippers/logstash.py
+++ b/shippers/logstash.py
@@ -53,6 +53,9 @@ class LogstashShipper:
         self._session.mount(self._logstash_url, HTTPAdapter(max_retries=retry_strategy))
 
     def send(self, event: dict[str, Any]) -> str:
+        if "_id" not in event and self._event_id_generator is not None:
+            event["_id"] = self._event_id_generator(event)
+
         self._events_batch.append(event)
         if len(self._events_batch) < self._max_batch_size:
             return _EVENT_BUFFERED
@@ -86,4 +89,4 @@ class LogstashShipper:
             )
             if self._replay_handler is not None:
                 for event in events:
-                    self._replay_handler("logstash shipper", self._replay_args, event)
+                    self._replay_handler("logstash", self._replay_args, event)
--- a/shippers/logstash.py
+++ b/shippers/logstash.py
@@ -7,11 +7,23 @@ from typing import Any, Optional
 
 import requests
 import ujson
+from requests.adapters import HTTPAdapter
+from requests.exceptions import RequestException
+from urllib3.util.retry import Retry
 
+from share import shared_logger
 from shippers.shipper import EventIdGeneratorCallable, ReplayHandlerCallable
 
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -5,5 +5,5 @@ elasticsearch>=7.16,<7.17
 PyYAML>=6.0,<6.1
 aws_lambda_typing>=2.16.1,<3
 ujson>=5.5.0,<5.6
-requests>=2.25.1
-
+requests<2.26,>=2.20.0
```
aspacca, line 8: particular reasons to bump down min version from `>=2.25.1` to  `>=2.20.0`?

I had a strategy in mind when pinning versions in `requirements.txt` that was to have the latest patch available (helped by the dependantbot) and lock to the next minor.

it's the best I could find to keep up without stale versions and still be quite confident that every builds use more or less the same version without explicitly pinning it.
maybe having `==X.Y.Z` would be as good or more, if we rely on the dependantbot

shippers/logstash.py
```diff
@@ -57,11 +76,17 @@ def flush(self) -> None:
...
-            # TODO: Change with actual handling
-            raise RuntimeError(f"Errors while sending data to Logstash. Return code {response.status_code}")
+        try:
+            self._session.put(
+                logstash_url,
+                data=gzip.compress(ndjson.encode("utf-8"), compression_level),
+                headers={"Content-Encoding": "gzip", "Content-Type": "application/x-ndjson"},
+                timeout=_TIMEOUT,
+            )
+        except RequestException as e:
```
aspacca, line 86: `RequestException` is raised only after fulfilling the retry strategy, is it?

tests/shippers/test_logstash.py
```diff
@@ -42,31 +46,72 @@
...
+    @responses.activate
     def test_flush(self) -> None:
-        logstash_shipper = LogstashShipper(logstash_url="http://logstash_url", max_batch_size=2)
+        url = "http://logstash_url"
+        responses.put(url=url, status=200)
+        responses.put(url=url, status=200)
+        logstash_shipper = LogstashShipper(logstash_url=url, max_batch_size=2)
         logstash_shipper.send(_dummy_event)
         assert logstash_shipper._events_batch == [_dummy_event]
         logstash_shipper.flush()
```
aspacca, line 117: do we miss a last assert on `logstash_shipper._events_batch == []`?

aspacca (commented):
nice! :)

shippers/logstash.py
girodav: If Logstash responds back with any of the status codes in `_STATUS_FORCE_LIST `, yes, the request will be retried and we handle the exception if the max number of retries is exceeded. Otherwise, the request is not retried and we fail fast (e.g if we get a 401 authorization denied)

requirements.txt
girodav: That came from localstack requirements

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
localstack 1.2.0 requires requests<2.26,>=2.20.0, but you have requests 2.26.0 which is incompatible.
localstack-ext 1.2.0 requires requests<2.26,>=2.20.0, but you have requests 2.26.0 which is incompatible.

```

but you are right, makes more sense to just have `==2.25.1`, which is the latest version before 2.26

tests/shippers/test_logstash.py
girodav: Line below (73) :)

requirements.txt
girodav: Changed in https://github.com/elastic/elastic-serverless-forwarder/pull/183/commits/7344129a6cbf5007e56d65645d2e863e258fcc2b

tests/shippers/test_logstash.py
aspacca: lol damn github folding code :D

aspacca (approved):

PR #185, opened 2022-11-22
[DOCS] Cut back README to summary and link to official docs

ollyhowell:
## What does this PR do?

Cut back README to short summary and link to official docs

## Why is it important?

Single source of docs will be https://www.elastic.co/guide/en/observability/master/aws-elastic-serverless-forwarder.html

DO NOT MERGE until this content is live via https://github.com/elastic/observability-docs/pull/2214

## Checklist

- [ ] ~My code follows the style guidelines of this project~
- [ ] ~I have commented my code, particularly in hard-to-understand areas~
- [ ] ~I have made corresponding change to the default configuration files~
- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [ ] ~I have added an entry in `CHANGELOG.md`~

## Author's Checklist

- [ ] Review and compare to https://github.com/elastic/observability-docs/pull/2214
- [ ] Merge only when above PR is merge in and new doc pages live

## Related issues
https://github.com/elastic/observability-docs/pull/2214


docs/README-AWS.md
```diff
@@ -1,829 +1,20 @@
...
-- [Using tags](#using-tags)
-- [Defining include/exclude filters](#defining-includeexclude-filters)
-- [JSON content discovery](#json-content-discovery)
-- [Expanding events from list in json object](#expanding-events-from-list-in-json-object)
-- [Manage multiline messages](#manage-multiline-messages)
-- [Routing AWS Service Logs](#routing-aws-service-logs)
-- [Setting up S3 event notifications for SQS](#setting-up-s3-event-notifications-for-sqs)
-- [Troubleshooting and error handling](#troubleshooting-and-error-handling)
-- [Resources and links](#resources-and-links)
+Please refer to the official [Elastic documentation for Elastic Serverless Forwarder](https://www.elastic.co/guide/en/observability/master/aws-elastic-serverless-forwarder.html) for detailed instructions on how to deploy and configure the forwarder.
```
aspacca, line 5: > https://www.elastic.co/guide/en/observability/master/aws-elastic-serverless-forwarder.html

When we'll release 1.6.0 of the forwarder can we pin to something like https://www.elastic.co/guide/en/observability/1.6.0/aws-elastic-serverless-forwarder.html?

If I got it right https://www.elastic.co/guide/en/observability/%VERSION% is the stack version, is it?

I find it troublesome in such case.
We'll push updated documentation for ESF 1.6.0 and probably backport to every stack versions the forwarder is compatibile with.
Where will the docs for ESF 1.5.0 end up? lost?

is https://www.elastic.co/guide/en/observability/master/aws-elastic-serverless-forwarder.html the docs for the "main" (master) branch of ESF or ESF compatible with master branch of the stack?
In the second case there will be multiple ESF versions compatibile, which ESF version will be the one effectively published at that URL?

How do I keep consistent the link across README sent to SAR for a specific ESF version to the documentation of that specific ESF version on https://www.elastic.co/guide?

I can see it can be done only if for https://www.elastic.co/guide/en/observability/%VERSION%, %VERSION% refers to the ESF version

ollyhowell: > If I got it right https://www.elastic.co/guide/en/observability/%VERSION% is the stack version, is it?

Yes, that is stack version.
Hmm, interesting problem (and not unique, though I don't know how well our systems are currently set for such a use case...). Some questions to work out the way forward:

1. Do you see users staying on older versions of ESF for a period of time or will they generally upgrade as soon as new version is available? How will they know a new version is released?
2. Do we actually need to keep the ESF docs versioned independently? i.e. keep version 1.5 readily available for compairson and for users even when 1.6 is released? Obviously previous versions will be available via GitHub source but not as nicely/usable as choosing a version from main docs
3. Would a link to the GitHub branch for that version of the ESF docs suffice? Or a tagged version via GitHub?

Generally I've solved this problem in the past by having packages with independent versioning being pulled into a super-docset that has its own versioning. From within the equivalent of main or 8.2 etc. you can see the latest version of the package/forwarder and then independently choose between 'sub' versions i.e. 1.5 of the ESF docs.

I can discuss with my team but we might have to consider NOT cutting back the docs in this case and keeping the README docs as independently versioned within this repo, and only publishing the latest version via /guide. It's not ideal and we might still have to use branches or tags within this repo to indicate the correct version to SAR.

Let me know what you think to questions and I will find out a bit more about what is possible or how we could build towards a more elegant solution...

ollyhowell:
Had a quick chat with @tommyers-elastic and adding some notes:

- simplest option might be to create a new book for ESF so we can retain independent versioning i.e. keep a placeholder page within Obs Guide but move the new content into its own like e.g. https://www.elastic.co/guide/en/elasticsearch/client/curator/current/index.html. This would get a /current and a /1.5, /1.6 etc.
- we can then include a compatibility matrix and easily document differences between versions and make them available to users
- for next release of ESF, need to investigate and update docs link so that docs within SAR point to correct place
- does it still make sense to merge this PR as it will link to updated docs from repo (and from SAR once new version is pushed)?

Will run this by @elastic/obs-docs team in case there is a better way and if not, will create an issue for creating a new book for the ESF docs.

ollyhowell:
Update: sounds like separate book is the way to go to retain control over versioning.
Will create new issue to track this part of the work: https://github.com/elastic/observability-docs/issues/2426

aspacca:
@ollyhowell I think we can merge the current status until we wait for a separate book, what do you think?

ollyhowell:
Buon Anno @aspacca!
Yes, I think we should merge this - setting to Ready for Review and if you can Approve then I can merge in (or feel free to merge yourself).

aspacca (approved):

PR #186, opened 2022-11-22
Logstash output integration tests

endorama:
## What does this PR do?

Adds integrations tests for Logstash output.

## Why is it important?

This is part of our work on Logstash output.

## Checklist

## Author's Checklist

- [ ]

- install all requirements from `requirements.txt`
- Run `py.test -vv -s tests/handlers/aws/test_outputs.py`

## Related issues

-


tests/handlers/aws/testdata/cloudwatch-log-1.json
```diff
@@ -0,0 +1,5 @@
+{
```
aspacca, line 1: total personal preference: if the payload it's not too big I prefer to have it as variable in the test, without having to skip back and forward between two files.

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,190 @@
...
+            CloudwatchLogStreamARN=cloudwatch_group_arn,
+            LogstashURL=self.logstash.get_url()
+        ))
+        print(self.config)
+
+        _upload_content_to_bucket(
+            client=self.s3_client,
+            content=self.config,
+            content_type="text/plain",
+            bucket_name=f"{type(self).__name__}-config-bucket",
```
aspacca, line 155: 👍

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,190 @@
...
+def load_file_fixture(name:str) -> str:
+    filepath = os.path.join(os.path.dirname(__file__), 'testdata', name)
+
+    res = ""
+    with open(filepath) as f:
+        res = f.read()
+
+    return res
+
+def _upload_content_to_bucket(client, content, key:str, bucket_name:str, content_type:str, acl:str="public-read-write") -> None:
```
aspacca, line 33: I see this is still a draft: in order to target main or the feature branch I would aim for having a `test_integration_utils.py` with this and all the other helpers below without duplication of what we already have around with similar functionality.

indeed this could be a pre-emptive separate PR: let's separate all the helpers in `test_integration_utils.py`, replace the current usage, and rely on new helpers from then onwards.

such PR can go directly to main

endorama: Indeed! My goal here was to understand how the code was working. One key difference between this version and the one in the other file is that this function takes `client` as an argument, instead of relying on `_mockaws_client`. This avoids client-reuse issues (as you control which client is being used).

tests/handlers/aws/testdata/config.yaml
```diff
@@ -0,0 +1,9 @@
+inputs:
```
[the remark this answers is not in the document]
endorama: I'm not sure if nesting folder is useful but for sure this file needs to be renamed.

Do you foresee a lot of files in there? If not I would just use a flat folder at the moment, and evaluate how to structure it after the refactoring.

tests/handlers/aws/testdata/cloudwatch-log-1.json
endorama: The idea behind moving that content to file is that is it possible to share some common fixtures. Doing that with variables would move them away from where they are used so I don't think is the right approach.

Reusing the same fixtures would help adding tests and normalizing input data across the test suite (I extracted three of them, I've no idea how many of there there are in the current tests)

tests/handlers/aws/test_outputs.py
girodav: One note, I'd consider adding a `lower()` call somewhere since bucket names are strictly lowercase (ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html). Things work now but I wouldn't be surprised if newer versions of localstack would enforce this.

tests/handlers/aws/test_outputs.py
endorama: I extracted the functions in a separate file in [2b031f6](https://github.com/elastic/elastic-serverless-forwarder/pull/186/commits/2b031f68d3523c800947201068a6c4ef793748b2)

tests/handlers/aws/test_outputs.py
endorama: Created a dedicated helper and used `lower` in this case, thanks for catching it!

tests/handlers/aws/testdata/cloudwatch-log-1.json
aspacca: Indeed I would not focus on reusing the same fixtures: if we did so far it was probably because of laziness ;)

The more fuzziness we could add to the fixtures the better it would be in my opinion: given the experience of the unit tests for storage, having the content totally random, helped to catch corner case bugs we would not have spotted otherwise.

but I'm find with your current solution for the moment :)

tests/handlers/aws/testdata/config.yaml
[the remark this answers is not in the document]
aspacca: Let's keep it for later

I think it's a lot a personal preference and what we find more readable: I prefer inline literal content without template, more local to where it is used, as long as it is of limited size. There's no right or wrong in my opinion

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
aspacca, line 1: minor. I would rename to `test_logstash.py` or something similar

I know that @girodav plans to split further `test_handler.py` so the more specific we are with the filename the more we can split giving the right hint just looking at the filename :)

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+class TestLambdaHandlerLogstashOutputSuccess(TestCase):
+    def setUp(self) -> None:
+        lst = LocalStackContainer(image="localstack/localstack:1.1.0")
+        lst.with_env("EAGER_SERVICE_LOADING", "1")
+        lst.with_services("s3", "logs", "sqs")
+        self.localstack = lst.start()
+
+        aws_default_region = "us-east-1"
+        session = boto3.Session(region_name=aws_default_region)
+        self.s3_client = session.client("s3", endpoint_url=self.localstack.get_url())
```
aspacca, line 40: this is very tidy 👍

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+        self.logs_client = session.client("logs", endpoint_url=self.localstack.get_url())
+        self.sqs_client = session.client("sqs", endpoint_url=self.localstack.get_url())
+
+        self.logstash_http_port = 5043
+        lgc = LogstashContainer(port=self.logstash_http_port)
+        logstash_config = """\
```
aspacca, line 46: minor: over the codebase I think we tend to use python 3.6 f-strings rather than format()

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+                port => {logstash_http_port}
+                codec => json_lines
+              }}
+            }}
+
+            output {{ stdout {{ codec => json_lines }} }}
+            """.format(
+            logstash_http_port=self.logstash_http_port
+        )
+        print(logstash_config)
```
aspacca, line 58: debug leftover?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+        self.fixtures = {
+            "cw_log_1": _load_file_fixture("cloudwatch-log-1.json"),
+            "cw_log_2": _load_file_fixture("cloudwatch-log-2.json"),
+        }
+
+        group_name = _class_based_id(self, suffix="source-group")
+        stream_name = _class_based_id(self, suffix="source-stream")
+
+        _logs_create_cloudwatch_logs_group(self.logs_client, group_name=group_name)
+        g = _logs_create_cloudwatch_logs_stream(self.logs_client, group_name=group_name, stream_name=stream_name)
```
aspacca, line 71: minor: let's give a more verbose name to `g` if you like as well :)

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+        self.group_name = group_name
+        self.stream_name = stream_name
+
+        config_content = _load_file_fixture("config.yaml")
+        self.config = Template(config_content).substitute(
+            dict(CloudwatchLogStreamARN=cloudwatch_group_arn, LogstashURL=self.logstash.get_url())
+        )
+        print(self.config)
```
aspacca, line 88: debug leftover?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        _s3_upload_content_to_bucket(
+            client=self.s3_client,
+            content=self.config,
+            content_type="text/plain",
+            bucket_name=_class_based_id(self, suffix="config-bucket").lower(),
+            key="folder/config.yaml",
+        )
+
+        def _create_sqs_queue(client: BotoBaseClient, name: str) -> Any:
```
aspacca, line 98: any reason for having this function defined in the `setUp` instead that as an helper?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+            q = client.create_queue(QueueName=name)
```
aspacca, line 99: minor: let's give a more verbose name to `q` if you like as well :)

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+            print(q["QueueUrl"])
```
aspacca, line 100: debug leftover?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+            return q["QueueUrl"]
+
+        os.environ["AWS_DEFAULT_REGION"] = aws_default_region
+        os.environ["S3_CONFIG_FILE"] = f"s3://{type(self).__name__}-config-bucket/folder/config.yaml".lower()
+        os.environ["SQS_CONTINUE_URL"] = _create_sqs_queue(self.sqs_client, f"{type(self).__name__}-continuing")
```
aspacca, line 105: using `_class_based_id` here?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+            key="folder/config.yaml",
+        )
+
+        def _create_sqs_queue(client: BotoBaseClient, name: str) -> Any:
+            q = client.create_queue(QueueName=name)
+            print(q["QueueUrl"])
+            return q["QueueUrl"]
+
+        os.environ["AWS_DEFAULT_REGION"] = aws_default_region
+        os.environ["S3_CONFIG_FILE"] = f"s3://{type(self).__name__}-config-bucket/folder/config.yaml".lower()
```
aspacca, line 104: using `_class_based_id` here?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+        os.environ["SQS_CONTINUE_URL"] = _create_sqs_queue(self.sqs_client, f"{type(self).__name__}-continuing")
+        os.environ["SQS_REPLAY_URL"] = _create_sqs_queue(self.sqs_client, f"{type(self).__name__}-replay")
```
aspacca, line 106: using `_class_based_id` here?

tests/handlers/aws/test_outputs.py
```diff
@@ -0,0 +1,135 @@
...
+        self.logstash.stop()
+
+        for k, m in self.mocks.items():
+            m.stop()
+
+    def test_foo(self) -> None:
+        event_cloudwatch_logs, event_ids_cloudwatch_logs = _logs_retrieve_event_from_cloudwatch_logs(
+            self.logs_client, group_name=self.group_name, stream_name=self.stream_name
+        )
+        print(event_cloudwatch_logs, event_ids_cloudwatch_logs)
```
aspacca, line 127: debug leftover?

tests/testcontainers/logstash.py
```diff
@@ -0,0 +1,130 @@
...
+        You can override any value set here by calling <instance>.with_env(...) after initializing this class
+        """
+        self.with_env("LOG_LEVEL", os.environ.get("LOGSTASH_LOG_LEVEL", "warn"))
+        self.with_env("CONFIG_STRING", "input { stdin {} } output { stdout {} }")
+        self.with_env("XPACK_MONITORING_ENABLED", "false")
+
+    def get_url(self) -> str:
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(self.port)
+        return "http://{}:{}".format(host, port)
```
aspacca, line 78: same for f-strings vs `str.format()`

tests/testcontainers/logstash.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def get_apiurl(self) -> str:
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(self.api_port)
+        return "http://{}:{}".format(host, port)
```
aspacca, line 83: same for f-strings vs `str.format()`

tests/testcontainers/logstash.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @wait_container_is_ready(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)  # type: ignore
+    def _connect(self) -> None:
+        url = self.get_apiurl()
+        response = requests.get("{}/?pretty".format(url), timeout=1)
```
aspacca, line 88: same for f-strings vs `str.format()`

aspacca (approved):
just minor comments

tests/handlers/aws/test_outputs.py
endorama: yes and no. Is there a way to have log this at debug level for example? In my debugging this (and the other below) has been key points that helped me debug the test so I think there is value having them here, but I would like to use something different than `print`. 

Is this done in other areas of the code?

tests/handlers/aws/test_outputs.py
endorama: Forgot about moving it, will do

tests/handlers/aws/testdata/cloudwatch-log-1.json
endorama: My understanding is that testing and fuzzing have different scope and should happen at different times. 

We need tests (unit/integration) to be reliable, predictable and easy to reproduce; that may not happen with random inputs or inputs that change through test runs.

On the other end we need fuzzing to catch errors with data that we cannot foresee or that we have not thought about.

Instead of trying to do both with our current fixtures, what about implementing fuzz test as separate tests?

tests/handlers/aws/test_outputs.py
aspacca: not tested, something like this might work:
```
os.environ["LOG_LEVEL"] = "DEBUG"

from share import shared_logger
shared_logger.debug(...)
```

if I'm not wrong setting the env variable before the shared_logger import is relevant

aspacca: > Is this done in other areas of the code?

it is done, but not in the tests

tests/handlers/aws/testdata/cloudwatch-log-1.json
aspacca: > Instead of trying to do both with our current fixtures, what about implementing fuzz test as separate tests?

if we do both, separated, I'm oki, but we should try to add them together, or at least have both by the time a feature is released.

if we do only one of that, I would prefer to have fuzzing over easy to reproduce tests with fixed fixtures: while the first are harder to maintain, they give broader coverage to corner cases and I prefer the trade off of an higher effort on the development side with the benefit of catching as much bugs as possible.

As long as I can understand clearly what test with what fuzziness init parameters is failing, and I have a way to replay the same random content.

The last is not 100% true at the moment: when a storage unit test failed because of the random content case we didn't manage properly, I had to add some code in a local branch to dump the random content to a file, and then run the test again until the same content case will occur. After that I have the content that gives a reliable and predictable test.

Yes, it's not easy to reproduce: 
> I had to add some code in a local branch to dump the random content to a file

this should be done by default, without needing to add code back and forth, and I should have at the same time the way to run a test telling if loading a dumped content or not. those artifacts should be stored by the CI job and I should be able to download them easily and feed them easily to the test.

if we have the above, and all this information is easy to access and be replayed, and it's easy to spot the failing test etc, and the test is not flaky by itself and there is a real bug in the tested code, I personally consider it "to be reliable, predictable and easy to reproduce" :)

The effort we should avoid is about understanding if a test run or not, if it failed, with what parameters, if we can trust the test in the first place, etc etc ;)

tests/handlers/aws/test_outputs.py
endorama: I will remove them, the debugger may help in such cases.

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,126 @@
...
+            m.start()
+
+    def tearDown(self) -> None:
+        self.localstack.stop()
+        self.logstash.stop()
+
+        for k, m in self.mocks.items():
+            m.stop()
+
+    def test_foo(self) -> None:
```
girodav, line 116: Maybe we could finally give a name to this test, now that everything is so tidy :)

girodav (approved):

PR #194, opened 2022-12-15
Add logstash output test failure case

endorama:
## What does this PR do?

This PR extracts Forwarder config loading to a separate function to enable creating a different test. It also adds a failing test case.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-


endorama:
@aspacca may you check the test checks the correct behaviour? I was expecting an exception to be triggered but it does not.

endorama:
(sorry for the multiple fix commit, my lint setup broke and I had to rely on the CI)

girodav:
> @aspacca may you check the test checks the correct behaviour? I was expecting an exception to be triggered but it does not.

what exception were you expecting? In the generic failure scenario, the forwarder should handle the exception and push the failed events into the replay queue.

endorama:
@girodav I think I misjudged this, as I was expecting the handler to emit an error. But the behaviour is not that, so the test is missing the check on the replay queue, which I'm adding. Sorry for the confusion!

tests/handlers/aws/test_logstash_output.py
```diff
@@ -128,3 +137,32 @@ def test_sent_messages(self) -> None:
...
+        handler(event_cloudwatch_logs, ctx)  # type: ignore
+        # test new input => output to stdout
+
+        msgs = self.logstash.get_messages()
+        assert len(msgs) == 0
+
+        sqs = self.aws_session.resource("sqs", endpoint_url=self.localstack.get_url())
+        queue = sqs.Queue(os.environ["SQS_REPLAY_URL"])
+        count = 0
+        for message in queue.receive_messages():
```
girodav, line 163: Based on [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message), we should set "MaxNumberOfMessages" to a number, as the default is 1. Alternatively we could change that `for` with a `while`. Otherwise we would get only one message returned by the API.

tests/handlers/aws/test_logstash_output.py
```diff
@@ -128,3 +137,32 @@ def test_sent_messages(self) -> None:
...
+            dict(CloudwatchLogStreamARN=self.cloudwatch_group_arn, LogstashURL="http://fake.url"),
+            "folder/config2.yaml",
+        )
+
+        event_cloudwatch_logs, event_ids_cloudwatch_logs = _logs_retrieve_event_from_cloudwatch_logs(
+            self.logs_client, group_name=self.group_name, stream_name=self.stream_name
+        )
+
+        ctx = ContextMock(TIMEOUT_15m)
+        handler(event_cloudwatch_logs, ctx)  # type: ignore
```
aspacca, line 154: please. let's keep the assertion on the expected output of the handler

in general I see that in the logstash integration tests we are reducing the number of assertions

aspacca:
we are missing testing the the continuing queue mechanism is working properly with logstash as output

we are also missing testing using two outputs (es and logstash) at the same time

:)

tests/handlers/aws/test_logstash_output.py
endorama: I added assertion to check message content, not just number of received messages.

girodav:
> @girodav May you please double check that assertion in [`d19df48` (#194)](https://github.com/elastic/elastic-serverless-forwarder/pull/194/commits/d19df4869eff2cfcfa3028e2e8ae8ff1afc4eace) are correct? I find strange that there are 2 messages to process and the first one is sent to Logstash and to the continuing queue as well, but I'm not sure if this is the correct behaviour or a bug.

The continuing queue handling is essentially the same as for ES.

We check how much time we have left only after attempting to send an event. If we realize that there is not enough time to continue processing events, we 1 - flush the current batch to events to ES/Logstash 2- publish the events to the continuing SQS queue. 

In this specific case, what happens is the following:

- First event is handled. It is gets and not actually sent to Logstash (the default batch size is 500)
- remaining time (0) is lower than the grace period (120s), we enter in the continuing queue handling logic
- Buffer is flushed, the first event is sent to Logstash
- Payload of events is pushed to the SQS continuing queue

tests/handlers/aws/test_logstash_output.py
aspacca: I would add full assertions on every field of the payload: `aws`, `cloud`, `log` and `message`

tests/handlers/aws/test_logstash_output.py
```diff
@@ -118,13 +123,77 @@ def tearDown(self) -> None:
...
+        ctx = ContextMock(0)
+        result = handler(event_cloudwatch_logs, ctx)  # type: ignore
+        assert result == "continuing"
+
+        msgs = self.logstash.get_messages()
+        assert len(msgs) == 1
+        assert msgs[0]["fields"]["message"] == self.fixtures["cw_log_1"].rstrip("\n")
+
+        messages = _sqs_get_messages(self.sqs_client, os.environ["SQS_CONTINUE_URL"])
+        assert messages[0]["Body"] == self.fixtures["cw_log_1"] + self.fixtures["cw_log_2"]
```
aspacca, line 199: I would trigger the lambda again and assure now only `self.fixtures["cw_log_2"]` will be sent to logstash, meaning the continuation from the last processed offset is working properly

endorama:
@aspacca I added 2 tests to check
1. the message content
2. the continuing queue offset processing

tests/handlers/aws/test_logstash_output.py
```diff
@@ -188,12 +191,91 @@ def test_send_timeout(self) -> None:
...
+        assert msgs[0]["fields"]["aws"]["cloudwatch"]["log_stream"] == stream_name
+        assert msgs[0]["fields"]["cloud"]["provider"] == "aws"
+        assert msgs[0]["fields"]["cloud"]["region"] == os.environ["AWS_DEFAULT_REGION"]
+        assert msgs[0]["fields"]["cloud"]["account"]["id"] == "000000000000"
+        assert msgs[0]["fields"]["log"]["offset"] == 0
+        assert msgs[0]["fields"]["log"]["file"]["path"] == f"{group_name}/{stream_name}"
+        assert msgs[0]["fields"]["message"] == self.fixtures["cw_log_1"].rstrip("\n")
+
+        assert msgs[1]["fields"]["aws"]["cloudwatch"]["log_group"] == group_name
+        assert msgs[1]["fields"]["aws"]["cloudwatch"]["log_stream"] == stream_name
```
aspacca, line 275: missing `assert msgs[1]["fields"]["aws"]["cloudwatch"]["event_id"] == event_ids_cloudwatch_logs[1]` :)

aspacca (commented):
nit: missing checking `aws.cloudwatch.event_id`

girodav (approved):

PR #210, opened 2023-01-13
Logstash support - technical preview

girodav:
## What does this PR do?

This PR introduces:
* First release of a new shipper, that enables ESF to send logs to a Logstash instance. It is intended to be a "Technology Preview".
* Unit tests and integration tests covering the most important basic scenarios.
* A new architecture diagram, listing Logstash as new possible output.

## Why is it important?

Refer to the master issue https://github.com/elastic/obs-infraobs-team/issues/809

## Checklist

## Author's Checklist

- [ ]

* Execute integration tests
` py.test -vv -s tests/handlers/aws/test_logstash_output.py`

## How to test this PR in AWS

* Install Logstash on an EC2 instance. See the official documentation https://www.elastic.co/guide/en/logstash/current/installing-logstash.html 
* Install [ESF Logstash plugin](https://rubygems.org/gems/logstash-input-elastic_serverless_forwarder)
`$LOGSTASH_FOLDER/bin/logstash-plugin install logstash-input-elastic_serverless_forwarder`
* Follow [ESF Logstash plugin docs](https://github.com/logstash-plugins/logstash-input-elastic_serverless_forwarder/blob/main/docs/index.asciidoc) to configure the plugin
* Deploy ESF on your AWS account. Refer to the following example configuration
```yaml
inputs:
  - type: "s3-sqs" #or another input type
    id: "arn:aws:sqs:%REGION%:%ACCOUNT%:%QUEUENAME%"
    outputs:
      - type: "logstash"
        args:
          logstash_url: logstash_url
          username: "username" #optional
          password: "password" #optional
          max_batch_size: 500 #optional
          compression_level: 1 #optional
          ssl_assert_fingerprint: "22:F7:FB:84:1D:43:3E:E7:BB:F9:72:F3:D8:97:AD:7C:86:E3:08:42" #optional
```

## Example event as received by Logstash

```
{
    "aws" => {
	    "s3" => {
		    "bucket" => {
			    "name" => "davide-girardi-vpcflowlogs-test",
			    "arn" => "arn:aws:s3:::davide-girardi-vpcflowlogs-test"
		    },
		    "object" => {
			    "key" => "AWSLogs/11111111111/vpcflowlogs/eu-west-1/2023/01/13/1111111111_vpcflowlogs_eu-west-1_fl-0cea9cbf050c152d5_20230113T1610Z_25f82e62.log.gz"
		    }
	    }
    },
    "message" => "2 1111111111 eni-0e854ebb664e54fb7 172.31.37.100 179.60.147.157 22 4954 6 8 480 1673626211 1673626264 ACCEPT OK",
    "cloud" => {
	    "region" => "eu-west-1",
	    "provider" => "aws",
	    "account" => {
		    "id" => "1111111111111111"
	    }
    },
    "log" => {
	    "file" => {
		    "path" => "https://davide-girardi-vpcflowlogs-test.s3.eu-west-1.amazonaws.com/AWSLogs/111111111111/vpcflowlogs/eu-west-1/2023/01/13/11111111111111111_vpcflowlogs_eu-west-1_fl-0cea9cbf050c152d5_20230113T1610Z_25f82e62.log.gz"
	    },
	    "offset" => 2529
    },
    "@timestamp" => 2023 - 01 - 17 T08: 58: 54.995298 Z,
    "@version" => "1",
    "_id" => "timestamp-hash-offset"
}
```

## Related issues

- https://github.com/elastic/obs-infraobs-team/issues/870


requirements.txt
```diff
@@ -5,3 +5,7 @@ elasticsearch==7.16.3
 PyYAML==6.0
 aws_lambda_typing==2.16.3
 ujson==5.6.0
+requests==2.25.1
+responses>=0.22.0
```
aspacca, line 9: I guess that `responses`, `testcontainers` and `amazon-kclpy` should go to `requirements-test.txt`

also https://github.com/elastic/elastic-serverless-forwarder/pull/210/files#diff-dd6813d8bd799e4b5e2af9f37e87fadc28004eecb26dc652ade372ce3e12d153R11 imports `urllib3`, we should add to `requirements.txt`

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,288 @@
...
+            self.logs_client, group_name=self.group_name, stream_name=self.stream_name
+        )
+
+        ctx = ContextMock(TIMEOUT_15m)
+        handler(event_cloudwatch_logs, ctx)  # type: ignore
+
+        msgs = self.logstash.get_messages()
+        assert len(msgs) == 2
+
+        assert msgs[0]["message"] == self.fixtures["cw_log_1"].rstrip("\n")
```
aspacca, line 149: I'd assert everything we expect in `msg[]`, like we do [here](https://github.com/elastic/elastic-serverless-forwarder/pull/210/files#diff-405288248598cf246486803f67aaf3389026f2e92321efdfa325efe6c42094daR268-R277)

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,288 @@
...
+        assert msgs[1]["message"] == self.fixtures["cw_log_2"].rstrip("\n")
```
aspacca, line 150: same as above

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,288 @@
...
+        handler(event_cloudwatch_logs, ctx)  # type: ignore
+
+        msgs = self.logstash.get_messages()
+        assert len(msgs) == 0
+
+        messages = _sqs_get_messages(self.sqs_client, os.environ["SQS_REPLAY_URL"])
+        assert len(messages) == 2
+
+        event1 = json_parser(messages[0]["Body"])
+        assert event1["event_payload"]["message"] == self.fixtures["cw_log_1"].rstrip("\n")
```
aspacca, line 175: I'd assert everything we expect in `event1["event_payload"]`, similarly like we do [here](https://github.com/elastic/elastic-serverless-forwarder/pull/210/files#diff-405288248598cf246486803f67aaf3389026f2e92321efdfa325efe6c42094daR268-R277)

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,288 @@
...
+
+        event2 = json_parser(messages[1]["Body"])
+        assert event2["event_payload"]["message"] == self.fixtures["cw_log_2"].rstrip("\n")
```
aspacca, line 178: same as above

tests/handlers/aws/test_logstash_output.py
```diff
@@ -0,0 +1,288 @@
...
+            dict(CloudwatchLogStreamARN=self.cloudwatch_group_arn, LogstashURL="http://fake.url"),
+            "folder/config2.yaml",
+        )
+
+        event_cloudwatch_logs, event_ids_cloudwatch_logs = _logs_retrieve_event_from_cloudwatch_logs(
+            self.logs_client, group_name=self.group_name, stream_name=self.stream_name
+        )
+
+        ctx = ContextMock(TIMEOUT_15m)
+        handler(event_cloudwatch_logs, ctx)  # type: ignore
```
aspacca, line 166: let's add an assertion on `handler` return value

tests/scripts/docker/mypy.sh
```diff
@@ -20,7 +20,7 @@ docker run \
   --rm lint_mypy \
   /bin/bash \
   -c "pip install --user -U pip
+      pip install --user -r requirements.txt --cache-dir ${docker_pip_cache}
```
aspacca, line 23: `pip install --user -r requirements.txt --cache-dir ${docker_pip_cache}`

having separated requirements file according to the scope, only executing the only one used for building the lambda we assure the exact intended versions dependencies are installed

if we change the order potentially an indirect dependency from the other two requirements file could change the final versions installed

there's probably a better way to insure this but I don't know about it

tests/scripts/docker/run_tests.sh
```diff
@@ -25,6 +25,6 @@ docker run \
   /bin/bash \
   -c "sudo service docker start
       pip install --user -U pip
-      pip install --user -r requirements-tests.txt --cache-dir ${docker_pip_cache}
       pip install --user -r requirements.txt --cache-dir ${docker_pip_cache}
+      pip install --user -r requirements-tests.txt --cache-dir ${docker_pip_cache}
```
aspacca, line 29: `pip install --user -r requirements.txt --cache-dir ${docker_pip_cache}`

having separated requirements file according to the scope, only executing the only one used for building the lambda we assure the exact intended versions dependencies are installed

if we change the order potentially an indirect dependency from the other two requirements file could change the final versions installed

there's probably a better way to insure this but I don't know about it

tests/share/test_config.py
```diff
@@ -20,6 +20,7 @@
     WhileMultiline,
     parse_config,
 )
+from share.config import LogstashOutput
```
aspacca, line 23: nit, we have just above
```
from share import (
    Config,
    CountMultiline,
    ElasticsearchOutput,
    IncludeExcludeFilter,
    IncludeExcludeRule,
    Input,
    Output,
    PatternMultiline,
    WhileMultiline,
    parse_config,
)
```

share should export `LogstashOutput`

tests/shippers/test_factory.py
```diff
@@ -6,7 +6,9 @@
 from unittest import TestCase
 
 from share import ElasticsearchOutput, Output
+from share.config import LogstashOutput
```
aspacca, line 9: nit, we have just above
```
from share import ElasticsearchOutput, Output
```
share should export LogstashOutput

tests/shippers/test_factory.py
```diff
@@ -6,7 +6,9 @@
...
 from shippers import ElasticsearchShipper, ProtocolShipper, ShipperFactory
+from shippers.logstash import LogstashShipper
```
aspacca, line 11: nit, we have just above
```
from shippers import ElasticsearchShipper, ProtocolShipper, ShipperFactory
```
shippers should export LogstashShipper

tests/shippers/test_factory.py
```diff
@@ -52,10 +54,32 @@ def test_create(self) -> None:
...
+
+        with self.subTest("create logstash shipper success with only logstash_url"):
+            shipper = ShipperFactory.create(
+                output_type="logstash",
+                logstash_url="http://myhost:8080",
+            )
+
+            assert isinstance(shipper, LogstashShipper)
+
+        with self.subTest("create logstash shipper success with only logstash_url"):
```
aspacca, line 66: it is not `with only logstash_url` :)

tests/shippers/test_logstash.py
```diff
@@ -0,0 +1,193 @@
...
+import pytest
+import responses
+import ujson
+from requests import PreparedRequest
+
+from shippers.logstash import _EVENT_SENT, _MAX_RETRIES, LogstashShipper
+
+_now = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+_dummy_event: dict[str, Any] = {
```
aspacca, line 25: not sure. everywhere we use `_dummy_event` without doing a `deepcopy` we are actually mutating it

tests/shippers/test_logstash.py
```diff
@@ -0,0 +1,193 @@
...
+            assert logstash_shipper.send(_dummy_event) == _EVENT_SENT
+        with self.subTest("Exceeds max retries, replay handler set"):
+            for i in range(_MAX_RETRIES):
+                responses.put(url=url, status=429)
+            responses.put(url=url, status=429)
+            logstash_shipper = LogstashShipper(logstash_url=url)
+            replay_handler = MagicMock(side_effect=_dummy_replay_handler)
+            logstash_shipper.set_replay_handler(replay_handler)
+            assert logstash_shipper.send(_dummy_event) == _EVENT_SENT
+            replay_handler.assert_called_once_with("logstash", {}, _dummy_event)
```
aspacca, line 112:
```
replay_handler.assert_called_once_with("logstash", {}, _dummy_event)
```
the second argument should contain the `output_args`, in this case i'd expect `logstash_url`

tests/shippers/test_logstash.py
```diff
@@ -0,0 +1,193 @@
...
+        with self.subTest("Exceeds max retries, replay handler not set"):
+            for i in range(_MAX_RETRIES):
+                responses.put(url=url, status=429)
+            responses.put(url=url, status=429)
+            replay_handler = MagicMock(side_effect=_dummy_replay_handler)
+            logstash_shipper = LogstashShipper(logstash_url=url)
+            assert logstash_shipper.send(_dummy_event) == _EVENT_SENT
+            replay_handler.assert_not_called()
+        with self.subTest("Authentication error, request is not retried"):
```
aspacca, line 121: that's tricky: the replay handler includes the `output_args`, so the `username` and `password`, but if they are wrong it's more likely because they were not properly set in the config file. if we send to the replay queue, the only way for those events to go through is to have in logstash the "wrong" credentials, and I see it's not what it's going to happen

but I wonder if having the replay queue empty is very specific to `401`. is there any other http code that will end up in not sending anything to the replay queue?

for me that should be rather an exceptional case. and indeed we could send also with `401`: once the user will set the replay queue as trigger of the lambda, the message will fails three times and end up in the DLQ.

not sending to the replay queue means those events are lost forever

tests/shippers/test_logstash.py
```diff
@@ -0,0 +1,193 @@
...
+            replay_handler = MagicMock(side_effect=_dummy_replay_handler)
+            logstash_shipper.set_replay_handler(replay_handler)
+            assert logstash_shipper.send(_dummy_event) == _EVENT_SENT
+            replay_handler.assert_called_once_with("logstash", {}, _dummy_event)
+
+    def test_send_https_ssl_fingerprint(self) -> None:
+        certpath = os.path.join(os.path.dirname(__file__), "ssl", "localhost.crt")
+        keypath = os.path.join(os.path.dirname(__file__), "ssl", "localhost.pkcs8.key")
+        server_address = ("localhost", 8080)
+        httpd = http.server.HTTPServer(server_address, http.server.SimpleHTTPRequestHandler)
```
aspacca, line 133: I'd rather use the real logstash from docker, in an integration test, also we should test where the case providing a non matching ssl fingerprint fill the replay queue,

tests/shippers/test_logstash.py
```diff
@@ -0,0 +1,193 @@
...
+            expected_event["_id"] = "_id"
+            assert _payload == [expected_event, expected_event]
+
+            return 200, {}, "okay"
+
+        def event_id_generator(event: dict[str, Any]) -> str:
+            return "_id"
+
+        url = "http://logstash_url"
+        responses.add_callback(responses.PUT, url, callback=request_callback)
```
aspacca, line 178: I'd rather use the real logstash from docker, in an integration test

tests/testcontainers/logstash.py
```diff
@@ -0,0 +1,146 @@
...
+              }}
+            }}
+
+            output {{ stdout {{ codec => json_lines }} }}
+            '''.format(logstash_http_port=self.logstash_http_port)
+        lgc.with_env("CONFIG_STRING", logstash_config)
+    """
+
+    _DEFAULT_IMAGE = "docker.elastic.co/logstash/logstash"
+    _DEFAULT_VERSION = "8.6.0"
```
aspacca, line 46: if we support minimum version `7.16` as for ES, let's use that

tests/shippers/test_logstash.py
girodav: `output_args` are not really used for Logstash, but I added it anyway as it can be helpful for users to troubleshoot errors ending up in the replay queue

tests/shippers/test_logstash.py
girodav: `_dummy_event` is only read, never modified, in these tests, unless you are afraid it can be modified by another test file

share/events.py
```diff
@@ -0,0 +1,23 @@
...
+
+def normalise_event(event_payload: dict[str, Any]) -> dict[str, Any]:
+    """
+    This method move fields payload in the event at root level and then removes it with meta payload
+    It has to be called as last step after any operation on the event payload just before sending to the cluster
+    """
+    if "fields" in event_payload:
+        fields: dict[str, Any] = event_payload["fields"]
+        for field_key in fields.keys():
+            event_payload[field_key] = fields[field_key]
```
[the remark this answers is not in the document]
girodav: Not sure I understand your concern. We are in control of how that `event_payload` is created, and what fields it contains, as its created separately in each trigger. `event_payload["fields"]["fields"]` cannot happen unless we deliberately decide to include it. Am I missing something?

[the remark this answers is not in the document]
aspacca: yes, sorry forget it: I've mixed up `fields` and `message`, and even in the `message` case from a json it will be a json and not the parsed dict

tests/shippers/test_logstash.py
aspacca: I remember having a case similar to [this](https://github.com/elastic/elastic-serverless-forwarder/pull/210/files#diff-cc6c179c69e1fc063e981ba448ac2f6c1afc88e270d5e275d59db1966cce98feR78-R79)
where assigning `_dummy_event` to a new variable without `deepcopy` and mutating the new variable mutated also `_dummy_event`

```
PyDev console: starting.
Python 3.9.16 (main, Dec  7 2022, 10:16:11) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
_dummy_event = {"this": "that"}
my_event = _dummy_event
my_event["that"] = "this"
print(my_event)
{'this': 'that', 'that': 'this'}
print(_dummy_event)
{'this': 'that', 'that': 'this'}
```

tests/shippers/test_logstash.py
aspacca: > `output_args` are not really used for Logstash

where the original output args like compression level etc come from if they are not in the `output_args` of the replay queue message?

tests/shippers/test_logstash.py
girodav: > that's tricky: the replay handler includes the output_args, so the username and password, but if they are wrong it's more likely because they were not properly set in the config file. if we send to the replay queue, the only way for those events to go through is to have in logstash the "wrong" credentials, and I see it's not what it's going to happen

With the current implementation, this is not correct. For Logstash, nothing is sent into `output_args`, and only `es_datastream_name` is sent for Elasticsearch outputs.
Regarding 401 errors, replayed events will go through as far as the config is fixed by the user, before replaying the events. This is true for both Elasticsearch and Logstash, from what I could see

> not sending to the replay queue means those events are lost forever
Apart from what mentioned above, I'd send failed events to the replay queue, regardless of the type of error.

tests/shippers/test_logstash.py
girodav: Yep, I agree. These were added when we did not have integration tests but they do not make a lot of sense now.

tests/shippers/test_logstash.py
girodav: same as above

aspacca (approved):

PR #211, opened 2023-01-17
Added support for tags and removed fields and meta entries from shipped events

girodav:


aspacca (approved):

shippers/es.py
```diff
@@ -206,7 +191,7 @@ def send(self, event: dict[str, Any]) -> str:
         if "_id" not in event and self._event_id_generator is not None:
             event["_id"] = self._event_id_generator(event)
 
-        self._normalise_event(event_payload=event)
+        event = normalise_event(event_payload=event)
```
aspacca, line 194: I don't see the file in `share` package with `normalise_event` function committed :)

Issue #212, opened 2023-01-24
Refactor _id generation


girodav:
As of today, we generate `_id` within Elastic Serverless Forwarder, with an "id generation function" defined for each input type (i.e cloudwatch, kinesis etc). 
The logic we follow for each input resembles the following:

- get unique ID specific to the input (cloudwatch event id, kinesis record sequence number...)
- get the first 10 hex chars of the SHA256 hash of the above
- pair it with the log offset

This logic helps with data integrity, ensuring that no duplicates are published in ES, and it proved to be effective in most cases. However, It has two important drawbacks:
- The pool of possible IDs is not big enough. Collisions may occur with few hundreds thousands of documents ingested on the same index. (more details as of why are available [here](https://en.wikipedia.org/wiki/Birthday_attack#Mathematics) )
- It creates extra burden for ElasticSearch, during indexing (more info [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html#_use_auto_generated_ids))

I propose the following two enhancements/refactoring
- **Short-term:** Refactor the ID generation functions to either take more bits from the SHA256 output or implement a different solution with a bigger pool of possible outputs.
- **Long-term/Needs evaluation** Disable the internal ID generation with a configuration parameter, and let ES auto-generate the id

girodav:
The `_id` affects the way Elastic Serverless Forwarder handles failures and the replay queue. Without it, it would not be feasible to know which specific events failed to be published to ES and handle the replay queue accordingly.
Due to the above, we will implement only the short term enhancement discussed above, for the time being.

Disabling the `_id` generation in its entirety needs further evaluation to see how to handle tradeoffs. In addition to this, it might not be worth adding this feature at all as there are options to remove the `_id` before indexing events in ES (e.g ingest pipeline)

PR #214, opened 2023-01-25
Changed ID generation logic to use AWS-provided ids

girodav:
## What does this PR do?

This PR changes the event ID generation logic for all inputs. The new logic combines different data provided by AWS (e.g S3 bucket-name, Kinesis Record Sequence Number, etc) to generate IDs. It removes any call to `hashlib.sha256` that is not really required and only adds further computational burden to the Lambda.
Integration tests were updated accordingly, to accommodate this change.

Docs will be updated as part of another PR: https://github.com/elastic/observability-docs/pull/2473 . However, a generic note about this change has been added to the README.

## Why is it important?

In its current implementation, the pool of possible IDs is too small to avoid duplicates. More details can be found at https://github.com/elastic/elastic-serverless-forwarder/issues/212

## Checklist

`make unit-test`
`make integration-test`

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Elastic Serverless Forwarder
 
 ## Important - v1.6.0
 
-#### Version 1.6.0 introduces a new event ID format which is backwards incompatible with previously indexed events. Be aware that previously indexed events would be duplicated if they triggered the forwarder again after upgrading to this version. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
+#### Version 1.6.0 introduces a new event ID format which is backwards incompatible with previously indexed events. Be aware that previously indexed events would be duplicated if they trigger the forwarder again after upgrading to this version. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
 
 ### Changelog [link](https://github.com/elastic/elastic-serverless-forwarder/blob/main/CHANGELOG.md)
 ### For AWS documentation, [go here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md)
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Elastic Serverless Forwarder
 
 ## Important - v1.6.0
 
-#### Version 1.6.0 introduces a new way to generate event IDs. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
+#### Version 1.6.0 introduces a new event ID format which is backwards incompatible with previously indexed events. Be aware that previously indexed events would be duplicated if they triggered the forwarder again after upgrading to this version. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
 
 ### Changelog [link](https://github.com/elastic/elastic-serverless-forwarder/blob/main/CHANGELOG.md)
 ### For AWS documentation, [go here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md)
--- a/README.md
+++ b/README.md
@@ -3,5 +3,9 @@
 # elastic-serverless-forwarder
 Elastic Serverless Forwarder
 
+## Important - v1.6.0
 
+#### Version 1.6.0 introduces a new way to generate event IDs. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
+
+### Changelog [link](https://github.com/elastic/elastic-serverless-forwarder/blob/main/CHANGELOG.md)
 ### For AWS documentation, [go here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md)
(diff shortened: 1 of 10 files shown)


handlers/aws/kinesis_trigger.py
```diff
@@ -31,12 +31,14 @@ def _handle_kinesis_continuation(
...
 
     sequence_number = kinesis_record["kinesis"]["sequenceNumber"]
+    partition_key = kinesis_record["kinesis"]["partitionKey"]
     stream_type, stream_name, _ = get_kinesis_stream_name_type_and_region_from_arn(event_input_id)
 
     message_attributes = {
         "config": {"StringValue": config_yaml, "DataType": "String"},
         "originalStreamType": {"StringValue": stream_type, "DataType": "String"},
         "originalStreamName": {"StringValue": stream_name, "DataType": "String"},
+        "originalPartitionKey": {"StringValue": partition_key, "DataType": "String"},
```
aspacca, line 41: minor: let's add `partion_key` at https://github.com/elastic/elastic-serverless-forwarder/blob/462c5dbba65ed7a76975dd9c30dce962c1eb4c9d/handlers/aws/kinesis_trigger.py#L70 as well

aspacca (approved):

README.md
```diff
@@ -3,5 +3,9 @@
 # elastic-serverless-forwarder
 Elastic Serverless Forwarder
 
+## Important - v1.6.0
 
+#### Version 1.6.0 introduces a new way to generate event IDs. More information is available at [our troubleshooting documentation](https://www.elastic.co/guide/en/observability/master/aws-serverless-troubleshooting.html).
```
tommyers-elastic, line 8: this suggests to  me there there is some configuration available to generate new IDs. i would rephrase with something more definite e.g. "introduces a new event ID format which is backwards incompatible with previously indexed events; be aware that previously indexed events would be duplicated if they triggered the forwarder again after upgrading to this version".

tommyers-elastic (approved):
change looks good - one comment on the docs.

handlers/aws/kinesis_trigger.py
```diff
@@ -122,6 +128,9 @@ def _handle_kinesis_record(
                             "name": stream_name,
                             "partition_key": kinesis_record["kinesis"]["partitionKey"],
                             "sequence_number": kinesis_record["kinesis"]["sequenceNumber"],
+                            "approximate_arrival_timestamp": int(
```
aspacca, line 131: let's put this in `meta.approximate_arrival_timestamp`, what do you think?

aspacca: I'd use milliseconds everywhere, since it's the greatest resolution we have across all inputs: 
```
int(float(kinesis_record["kinesis"]["approximateArrivalTimestamp"]) * 1000)
```

aspacca (commented):
see comments

girodav:
Last few changes, adding timestamps to the IDs, were made to comply with this issue https://github.com/elastic/integrations/issues/5100

README.md
girodav: Good point 👍 

I already mentioned your last point in the troubleshooting documentation, but I agree that is best re-iterating it here, as it's the first thing users would see.

tests/handlers/aws/test_handler.py
```diff
@@ -1764,57 +1769,73 @@ def test_lambda_handler_replay(self) -> None:
...
+        )
+
+        prefix_s3_first = f"{event_time}-{bucket_arn}-{first_filename}"
+        prefix_s3_second = f"{event_time}-{bucket_arn}-{second_filename}"
 
         _event_to_sqs_message(queue_attributes=self._queues_info["source-sqs-queue"], message_body=self._cloudwatch_log)
-        event_sqs = _event_from_sqs_message(queue_attributes=self._queues_info["source-sqs-queue"])
+        timestamp = str(int(datetime.datetime.utcnow().timestamp()))
+        queue_attributes = self._queues_info["source-sqs-queue"]
+        queue_attributes["sent_timestamp"] = timestamp
```
aspacca, line 1783: I'm curious why this is needed at all: does not `sqs_client.receive_message()` include `messages["Messages"][]["attributes"]["SentTimestamp"]` in the first place?

aspacca: minor: replacing
[`messages = sqs_client.receive_message(QueueUrl=queue_attributes["QueueUrl"], MessageAttributeNames=["All"])`](https://github.com/elastic/elastic-serverless-forwarder/blob/a588ec0995d9a04aab916c81d64693568bb0b7e4/tests/handlers/aws/test_handler.py#L1235) with `messages = sqs_client.receive_message(QueueUrl=queue_attributes["QueueUrl"], AttributeNames=["All"], MessageAttributeNames=["All"])` will prevent the need for `queue_attributes["sent_timestamp"] = timestamp`

We can then have `_event_from_sqs_message` return `event_sqs` and `event_sent_timestamp`, similarly to what we do with `_event_from_cloudwatch_logs`

aspacca (approved):
minor refactoring of the test could be considered, not necessarily in the current PR

tests/handlers/aws/test_handler.py
girodav: You are totally right. I did not think about that :).

README.md
```diff
@@ -3,5 +3,9 @@
 # elastic-serverless-forwarder
 Elastic Serverless Forwarder
 
+## Important - v1.6.0
```
aspacca, line 6: just realised this is not the readme published on SAR: rather the one at https://github.com/elastic/elastic-serverless-forwarder/blob/main/docs/README-AWS.md

PR #223, opened 2023-01-31
use GitHub actions

v1v:
## What does this PR do?

Use GitHub actions for the CI, running each `make docker-test*` individually

## Why is it important?

Jenkins will be decommissioned in the future

## Failures

`make docker-coverage` failed with

```
tests/storage/test_s3.py::test_get_by_lines[newline length 2 for content type _IS_MULTILINE_WHILE with length multiplier 102400 and `json_content_type` disabled] 
Error: Process completed with exit code 2.
```

https://github.com/elastic/elastic-serverless-forwarder/actions/runs/4066717667/jobs/7003169458#step:4:1982

No clue what's the error about

## Metrics

It takes ~45 minutes in Jenkins the whole build and a bit more in GH actions

image

Let's see how long it takes in GH actions


.github/workflows/test.yml
```diff
@@ -0,0 +1,52 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  lint:
```
aspacca, line 17: nit: the order of the jobs is the order they are executed? license was executed before test in jenkins

v1v: They run in parallel. To run sequentially each job will need a `needs` line with the previous job name

aspacca:
> `make docker-coverage` failed with
> 
> ```
> tests/storage/test_s3.py::test_get_by_lines[newline length 2 for content type _IS_MULTILINE_WHILE with length multiplier 102400 and `json_content_type` disabled] 
> Error: Process completed with exit code 2.
> ```
> 
> https://github.com/elastic/elastic-serverless-forwarder/actions/runs/4066717667/jobs/7003169458#step:4:1982
> 
> No clue what's the error about
> 
> ## Metrics
> It takes ~45 minutes in Jenkins the whole build and a bit more in GH actions

it seems the error you are facing comes from [`timeout 45m /bin/bash ./tests/scripts/run_tests.sh`](https://github.com/elastic/elastic-serverless-forwarder/blob/main/tests/scripts/docker/run_tests.sh#L30)

exit code 2 in pytest is user interrupted (because of `timeout`)

if the make target takes longer on github action, if possible I would prefer to increase the resource available in the action, if not we can raise the timeout :)

aspacca:
btw, `docker-coverage` is the equivalent of running `docker-integration-test` and `docker-unit-test` in the same pytest execution with the following env variables exported:
```
PYTEST_ADDOPTS=--cov=. --cov-context=test --cov-config=.coveragerc --cov-branch
COVERAGE_FILE=.coverage
```

not sure what happens if you run the two separated test target adding the env variables to the coverage outcome: each target has a partial coverage. I don't know if running the two pointing to the same coverage file will produce the total coverage

v1v:
> btw, docker-coverage is the equivalent of running docker-integration-test and docker-unit-test in the same pytest execution with the following env variables exported:

What do you suggest? running the two make goals with the env variables so the parallelisation helps with faster builds or run the `docker-coverage` with a higher timeout?

aspacca:
> What do you suggest? running the two make goals with the env variables so the parallelisation helps with faster builds or run the `docker-coverage` with a higher timeout?

if we can run the two make goals and still have the total coverage is preferable. I have to investigate if that's possible

maybe that's the default behaviour: you can try running the two goals only with the env variables, assuring the junit xml file is persisted between each run, and look at the coverage report. if it's still 100% for everything that's enough :)

v1v:
> maybe that's the default behaviour: you can try running the two goals only with the env variables, assuring the junit xml file is persisted between each run, and look at the coverage report. if it's still 100% for everything that's enough :)

I'm not sure if that's even possible. I'll re-enable the same behaviour as done in Jenkins, so we can revisit the improvement for running more things in parallel and keep the code coverage report

aspacca:
> I'm not sure if that's even possible

it should be possible with https://github.com/actions/download-artifact

but according to https://github.com/actions/upload-artifact, if you upload the artifact with the same name from different jobs the content will be overwritten by the last executed upload

you can have different upload name, and download them all. but then we have to problem to merge the junitxml file

aspacca:
> but then we have to problem to merge the junitxml file

maybe that's not needed since https://github.com/irongut/CodeCoverageSummary support multiple junitxml files

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  license:
+    strategy:
+      fail-fast: false
+      matrix:
+        goal:
+          - docker-lint
+          - license
```
v1v, line 23:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
...
+    - uses: actions/checkout@v3
+    - run: make license
+
+  lint:
+    strategy:
+      fail-fast: false
+      matrix:
+        goal:
+          - docker-lint
+          - license
```
v1v, line 36:
```suggestion
```

aspacca:
> I'm not sure if that's even possible. I'll re-enable the same behaviour as done in Jenkins, so we can revisit the improvement for running more things in parallel and keep the code coverage report

@v1v 
do you mind if I test on your branch to run the `docker-*-test` separately producing different junit files and see if https://github.com/irongut/CodeCoverageSummary#filename matching all of them will properly show the "sum" of the coverage?

v1v:
> do you mind if I test on your branch to

Thanks for asking, sure thing. You can freely contribute to my PRs anytime! 🙏

aspacca (approved):

aspacca:
@v1v #260 lowered `make docker-coverage` execution time on jenkins from 53 mins ca to 27 mins ca

with that it seems we can now merge this PR as it is :)

v1v:
@aspacca  , there is no support for `/test` at the moment a couple of reasons:
- GH actions are fully integrated within the GH PR view, so in a couple of clicks it's possible to rerun a failed job
- GH ecosystem seems a bit more reliable and well maintained. So flakiness might not happen often.

In addition, since GH actions use a different idiosyncrasy we have not enabled the GH command integration, though we can revisit this in a follow up if needed

If no concerns, I'll say we can merge this PR now

aspacca:
> there is no support for `/test` at the moment a couple of reasons:
> 
> * GH actions are fully integrated within the GH PR view, so in a couple of clicks it's possible to rerun a failed job

indeed, @v1v, I've realised it after trying the `/test` comment :)



> If no concerns, I'll say we can merge this PR now

good to go for me

PR #227, opened 2023-02-02
Added SHA384 hashing to event ID format

girodav:
## What does this PR do?

This PR changes the event ID format to use a SHA3_384 hash of the AWS-related information we use to generate the ID (e.g S3 bucket name, kinesis stream name, etc)

## Why is it important?

Some AWS resources allows to have very long names, sometimes equal to the maximum allowed ID length in Elasticsearch (e.g Cloudwatch allows log group names of 512 characters). Since we use this information to generate the ID, as is, we can generate IDs that are effectively bigger than 512 bytes (ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html)

## Checklist

## Related issues

- Relates to https://github.com/elastic/elastic-serverless-forwarder/issues/212


tests/handlers/aws/test_utils.py
```diff
@@ -0,0 +1,155 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
aspacca, line 1: 🥇

aspacca (approved):

PR #260, opened 2023-02-27
Refactor integration tests

aspacca:
Cleanup/Bug
## What does this PR do?
Refactor integration tests in order to let them bring up containers only once and speed up execution time.
We added as well some scenario coverage that lead to a couple of bug fix (specifically regarding input with multiple outputs)

## Why is it important?
Approach to integration tests started diverging on the effort made for coverage about logstash shipper.
This PR aim to capitalise on the new approach, filling the gap on the diversion between the "legacy" integration tests and the new ones.
Moreover, the premise of creating no fixed infra artifacts (sqs queues, kinesis data streams, etc) in the integration tests was pushed further so to have a single integration test class with proper setup and tear down (both at class and at test methods levels) in order to bring up the required docker containers only once for integration tests execution, reducing drastically the total amount of time required for bootstrapping.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-


handlers/aws/replay_trigger.py
```diff
@@ -15,19 +15,20 @@ class ReplayedEventReplayHandler:
     def __init__(self, replay_queue_arn: str):
         self._replay_queue_arn = replay_queue_arn
         self._failed_event_ids: list[str] = []
-        self._event_ids_with_receipt_handle: dict[str, str] = {}
+        self._events_with_receipt_handle: dict[str, str] = {}
```
aspacca, line 18: changes if this file address the first bugfix:
- proper handling of multiple outputs in the replay queue handler

the event id is the same across multiple outputs, when the replay handler of the replay queue applies its logic in order to understand if the events were successfully replied we must rely on a event _unique_ id suffixed by the output type

handlers/aws/utils.py
```diff
@@ -302,7 +301,7 @@ def get_trigger_type_and_config_source(event: dict[str, Any]) -> tuple[str, str]
                 and "output_args" in event_body
                 and "event_payload" in event_body
             ):
-                return "replay-sqs", CONFIG_FROM_PAYLOAD
+                return "replay-sqs", CONFIG_FROM_S3FILE
```
aspacca, line 304: it does not make really sense to load the config from the payload in the case of the replay queue.

the idea of the config being read from the s3 file or the payload in the continuing queue is given by the fact that a continuation happens when a trigger payload was not able to be fully forwarded: we somehow postpone finishing the execution of the trigger in the future, where the s3 config file might change but we want to ensure that we will still execute with the original one

but in the case of the replay queue we are reacting on a forwarding failure: this might be indeed because of "wrong" information (like credentials, or URLs) contained in the s3 config file, that lead the forwarding to fail in the first place.
the replay queue is a failure recovery scenario: having the replay queue as an enabled trigger of the forwarder lambda is an explicit intention of the users, if they are not given the option to upload a new "fixed" s3 config file we prevent them to apply the recovery conditions

shippers/composite.py
```diff
@@ -64,7 +65,10 @@ def send(self, event: dict[str, Any]) -> str:
             return EVENT_IS_FILTERED
 
         for shipper in self._shippers:
-            shipper.send(event)
+            # dict are mutated if not deep copied, every shipper can mutate the
```
aspacca, line 68: the changes in this file are related to the second bugfix:
- Fix events mutation across different outputs of the same input

`shipper.send()` is expected to mutate the `event` we pass to it: in the case the `CompositeShipper` composes multiple shippers, the mutation will spread across the shippers (like carrying metadata or similar specific to a shipper: see an example of the bug where the tag related to the dataset added by the es shipper is still hold on the logstash shipper event).

storage/decorator.py
```diff
@@ -412,12 +412,15 @@ def __call__(
...
             # in this case we could have a trailing new line in what's left in the buffer
-            # or the content had a leading `{` but was not a json object before the circuit breaker intercepted it:
+            # or the content had a leading `{` but was not a json object before the circuit breaker intercepted it
+            # or we waited for the object start and never reached:
             # let's fallback to by_lines()
             if not self._is_a_json_object:
-                for line, starting_offset, ending_offset, original_newline_length, _ in self._by_lines_fallback(
-                    self._unfinished_line
-                ):
+                buffer: bytes = self._unfinished_line
```
aspacca, line 419: the changes in this file address the third bugfix:
- proper handling of non json content where no json object start was ever met in the storage json collector

if we reached this if branch, the content was not identified as a json content, information hold by the flag `self._is_a_json_object`. This flag value start as `False`, and will be set to `True` only if at least one json object was able to be parsed from the storage content, that's correct

the if branch yields then `self._unfinished_line`, where all the previous storage content that was collected before the circuit breaker kicked in

the bug resides in the fact that that previous storage content is collected on `self._unfinished_line`, only if something that looked like a potential json object was ever met (ie: a line having a leading `{`).
we still have the case where this scenario never happened: this information is hold on the `wait_for_object_start` flag (where `False` means we finally found a potential json object) and the previous storage content that was collected so fare is instead in `wait_for_object_start_buffer`

tests/handlers/aws/utils.py
```diff
@@ -36,18 +54,19 @@ def _load_file_fixture(name: str) -> str:
     return res
 
 
-def _class_based_id(klass: object, prefix: str = "", suffix: str = "") -> str:
+def _time_based_id(prefix: str = "", suffix: str = "") -> str:
```
aspacca, line 57: refactoring this to being a time based rather than class/method based, allows us to execute once across different methods (like the setup and the effective test ones) and just needing to care for keeping the state of the result where it must be, in order to have unique and not colliding ids across multiple invocations

tests/testcontainers/es.py
```diff
@@ -0,0 +1,202 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
aspacca, line 1: this file contains porting of bootstrap code of the es container from the previous integration tests to testcontainers

on top of that there's some logic to handle state reset invoked by test method teardown

tests/testcontainers/logstash.py
```diff
@@ -23,128 +28,172 @@ class LogstashContainer(DockerContainer):  # type: ignore
         with LogstashContainer() as lc:
             # NOTE: container will terminate once out of this with statement
             url = lc.get_url()
-
-    With custom config
-    ------------------
-    ::
-
-        lgc = LogstashContainer(port=self.logstash_http_port)
```
aspacca, line 31: I removed this feature since I opted to always running logstash with ssl, that was the only the case in the previous logstash tests this feature was used.

I don't think we really need to have such level of customisation here, but I'm happy to discuss different opinions :)

tests/testcontainers/logstash.py
```diff
@@ -23,128 +28,172 @@ class LogstashContainer(DockerContainer):  # type: ignore
...
 
     def _configure(self) -> None:
         """
-        You can override any value set here by calling <instance>.with_env(...) after initializing this class
+        Values set here will override any value set by calling <instance>.with_env(...)
+        after initializing this class before <instance>.start()
         """
-        self.with_env("LOG_LEVEL", os.environ.get("LOGSTASH_LOG_LEVEL", "warn"))
+
+        # TODO: let's generate the certificate and key with code, current cert will expire Nov 17 15:04:09 2025 GMT
```
aspacca, line 72: logstash docker image misses the tool in order to generate the cert and key inside the container. I will look on how to do from code in python, and then save the output on files in the container and reference them instead

tests/testcontainers/logstash.py
```diff
@@ -23,128 +28,172 @@ class LogstashContainer(DockerContainer):  # type: ignore
...
+        self._last_reset_message_count = self._previous_message_count
+
+    def start(self) -> LogstashContainer:
+        self._configure()
         super().start()
         self._connect()
         return self
 
-    def get_messages(self, retry: int = 2, timeout: int = 1, delay: int = 1) -> list[dict[str, Any]]:
+    def get_messages(self, expected: int, retry: int = 2, wait_time: int = 1, delay: int = 1) -> list[dict[str, Any]]:
```
aspacca, line 146: I've introduced the `expected` param in order to stabilise the reliability of the method

the previous logic was floating, considering we now run a single container where every integration test is run on, and I had multiple random failure related to the assertion on the results of this method

aspacca:
`make docker-coverage` passed from lasting 53 minutes ca ([latest branch merged to master](https://beats-ci.elastic.co/blue/organizations/jenkins/Library%2Felastic-serverless-forwarder-mbp/detail/main/139/pipeline/111/)) to 27 minutes ca [in the current branch](https://beats-ci.elastic.co/blue/organizations/jenkins/Library%2Felastic-serverless-forwarder-mbp/detail/PR-260/5/pipeline/128)

tests/handlers/aws/test_handler.py
```diff
@@ -475,82 +444,30 @@ def test_lambda_handler_noop(self) -> None:
 
                 assert handler(lambda_event, ctx) == "exception raised: Exception('raised')"  # type:ignore
 
+        with self.subTest("raising unexpected exception apm client not Nome"):
```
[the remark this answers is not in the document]
aspacca: fixed

aspacca:
damn, seems to be flaky :(

aspacca:
I have the impression that flakiness is related to localstack

aspacca:
@endorama , @girodav this seems stable now

I'd merge it as soon as possible, thanks :)

tests/testcontainers/logstash.py
```diff
@@ -63,28 +70,76 @@ def __init__(
         self._last_reset_message_count: int = 0
         self._previous_message_count: int = 0
 
+    @staticmethod
+    def _certificates_for_authority_and_server(service_identity: str, key_size: int = 1024) -> tuple[str, str]:
```
girodav: I am fine with generating the cert in code, but what exactly are we gaining by adding this functionality in the tests?

girodav (approved):

tests/testcontainers/logstash.py
aspacca: that in two years the tests won't broke because of the expired cert and nobody will remember where to look :D

girodav: Fair enough lol

PR #276, opened 2023-03-08
Issue 264

### LINKED ISSUE #264: ThrottlingExceptions while processing data from cloudwatch input
Note: this is a placeholder issue; I will add more details soon.

***

Please include configurations and logs if available.

For confirmed bugs, please report:
- Version:
- Steps to Reproduce:

aspacca:
## What does this PR do?
Removes call to logs DescribeLogStreams api, building the potential log group and stream ARNs programmatically

## Why is it important?
DescribeLogStreams API is limited to 5 invocations per second: when concurrent execution of the forwarder are over that number the call get throttled and an exception is raised. We have no way to apply any throttling logic directly in the forwarder since we cannot know what is the proper wait/backoff value to use, because it is dependent to the current concurrent executions of the lambda, and there's no way to know them.
Instead we build a list of all the potential matching ARNs and check if an input is set.

## Checklist

## Related issues

- Closes #264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ##### Bug fixes
 * Fix events mutation across different outputs of the same input, proper handling of multiple outputs in the replay queue handler, proper handling of non json content where no json object start was ever met in the storage json collector [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/260)
 * Fix throttling exception in `cloudwatch-logs` input related to DescribeLogStreams [#276](https://github.com/elastic/elastic-serverless-forwarder/pull/276)
-*
+
 ### v1.7.1 - 2023/02/16
 ##### Bug fixes
 * Fix wrong resolved `expand_event_list_from_field` with AWS CloudTrail due to race condition [#244](https://github.com/elastic/elastic-serverless-forwarder/pull/244)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-### v1.7.2 - 2023/02/27
+### v1.7.2 - 2023/03/08
 ##### Bug fixes
-* Fix events mutation across different outputs of the same input, proper handling of multiple outputs in the replay queue handler, proper handling of non json content where no json object start was ever met in the storage json collector [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/260)
-
+* Fix events mutation across different outputs of the same input, proper handling of multiple outputs in the replay queue handler, proper handling of non json content where no json object start was ever met in the storage json collector [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/276)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
-### v1.7.2 - 2023/02/27
+### v1.7.2 - 2023/03/08
 ##### Bug fixes
-* Fix events mutation across different outputs of the same input, proper handling of multiple outputs in the replay queue handler, proper handling of non json content where no json object start was ever met in the storage json collector [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/260)
-
+* Fix events mutation across different outputs of the same input, proper handling of multiple outputs in the replay queue handler, proper handling of non json content where no json object start was ever met in the storage json collector [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/276)
+* Fix throttling exception in `cloudwatch-logs` input related to DescribeLogStreams [#260](https://github.com/elastic/elastic-serverless-forwarder/pull/276)
```
zmoog, line 4: Minor: fix the issue number in the link.

```suggestion
* Fix throttling exception in `cloudwatch-logs` input related to DescribeLogStreams [#276](https://github.com/elastic/elastic-serverless-forwarder/pull/276)
```

handlers/aws/utils.py
```diff
@@ -378,32 +370,63 @@ def get_account_id_from_arn(lambda_arn: str) -> str:
...
+    """
+    This function is not less resilient than the previous get_log_group_arn_and_region_from_log_group_name()
+    We avoid to call the describe_log_streams on the logs' client, since we have no way to apply the proper
+    throttling because we'd need to know the number of concurrent lambda running at the time of the call.
+    The list of regions is unluckily hardcoded: there's no API to get the list for logs (the most similar API is in ec2:
+    https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_regions.html - but we
+    would need to add IAM permissions for it).
+    We must keep the hardcoded list up to date or find a way to sync it automatically
+    """
+    for region in [
```
girodav, line 385: Personally, I'd rather use describe_regions instead https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_regions.html . We still rely on what cached by botocore, but it'd be easier to maintain.

girodav, line 385: I'd personally rather add the permissions to EC2:DescribeRegions instead of having to maintain this. We would stil have to rely on what cached in botocore, but would be easier to maintain.
Just my opinion, I wouldn't block the PR for this as they are both compromises and not optimal solutions.

zmoog (approved):

handlers/aws/utils.py
zmoog: I would consider adding `EC2:DescribeRegions` for ease of maintenance. 

But I believe we should ship this fix as soon as it is ready to users having this issue.

girodav (approved):

PR #280, opened 2023-03-08
action: use the host for running the coverage

v1v:
## What does this PR do?

Use the host for running the coverage so there is no need to run DinD

## Why is it important?

ITs are unstable when running DinD, so let's avoid using that

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,10 @@ jobs:
     - run: pip install -r requirements-tests.txt
     - run: pip install -r requirements.txt
     - run: make coverage
+      env:
+        # See https://github.com/elastic/elastic-serverless-forwarder/pull/280#issuecomment-1461554126
+        AWS_ACCESS_KEY_ID: AWS_ACCESS_KEY_ID
+        AWS_SECRET_ACCESS_KEY: AWS_SECRET_ACCESS_KEY
 
     - name: Store test results
       if: success() || failure()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,11 @@ jobs:
 
     - uses: actions/checkout@v3
 
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
     - run: make all-requirements
 
     - run: make coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,9 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - run: make all-requirements
+
+    - run: make coverage
 
     - name: Store test results
       if: success() || failure()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,10 +45,7 @@ jobs:
         python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
         cache: 'pip'          # caching pip dependencies
 
-    - run: make requirements
-    - run: make requirements-tests
-
-    - run: make coverage
+    - run: make requirements-test && make requirements && make coverage
 
     - name: Store test results
       if: success() || failure()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
         python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
         cache: 'pip'          # caching pip dependencies
 
-    - run: make requirements-test && make requirements && make coverage
+    - run: make requirements-tests && make requirements && make coverage
 
     - name: Store test results
       if: success() || failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -40,7 +40,14 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
+    - run: make all-requirements
```
v1v, line 48:
```suggestion
    - run: make requirements-tests
```

.github/workflows/test.yml
```diff
@@ -40,7 +40,14 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
+    - run: make requirements-tests
```
v1v, line 48:
```suggestion
    - run: make requirements
    - run: make requirements-tests
```

.github/workflows/test.yml
```diff
@@ -40,7 +40,15 @@ jobs:
...
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
+    - run: make requirements
+    - run: make requirements-tests
```
aspacca, line 49: please. execute `make requirements-tests` before `make requirements`: it's not optimal but this way we are sure that any sub-dependencies version from `requirements-tests` will not be "skewed" from how it should be from `requirements`

aspacca: btw, looking at the errors in the last run I have the impression that every `run` directive is executed in a different "context".
maybe the following?
```
run: make requirements-test && make requirements && make coverage
```

.github/workflows/test.yml
```diff
@@ -40,7 +40,15 @@ jobs:
...
+
+    - run: make coverage
```
v1v, line 51:
```suggestion
    - run: make requirements-test && make requirements && make coverage
```

aspacca: sorry, there's a typo: `requirements-tests`

.github/workflows/test.yml
```diff
@@ -40,7 +40,12 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
+    - run: make requirements-test && make requirements && make coverage
```
v1v, line 48:
```suggestion
    - run: make requirements-tests && make requirements && make coverage
```

.github/workflows/test.yml
v1v: As far as I see in the examples for using the [`setup-python`](https://github.com/actions/setup-python#caching-packages-dependencies)

```
steps:
- uses: actions/checkout@v3
- uses: actions/setup-python@v4
  with:
    python-version: '3.9'
    cache: 'pip'
    cache-dependency-path: '**/requirements-dev.txt'
- run: pip install -r subdirectory/requirements-dev.txt
```

Maybe the issue could be related to `pip3` used within the `make` goals

.github/workflows/test.yml
```diff
@@ -40,7 +40,12 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - run: make docker-coverage
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9' # As defined in tests/scripts/docker/run_tests.sh
+        cache: 'pip'          # caching pip dependencies
+
+    - run: make requirements-tests && make requirements && make coverage
```
v1v, line 48:
```suggestion
    - run: pip install -r requirements-tests.txt
    - run: pip install -r requirements.txt
    - run: make coverage
```

aspacca:
we need to set dummy env variables:
```
AWS_ACCESS_KEY_ID=AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY=AWS_SECRET_ACCESS_KEY
```

v1v:
> we need to set dummy env variables:

Do you mean within the make build system rather than the GitHub actions?

aspacca:
> Do you mean within the make build system rather than the GitHub actions?

in the GitHub actions only, it does not seem they are set or that the `run: make coverage` sees them.
the aws python client connects to localstack running on docker, but it expects some credentials. that we provide through the two above environment variables. they can have dummy value since localstack does not really care for them

aspacca:
>ModuleNotFoundError: No module named 'unittest.mock'
that's very unlikely: it is a builtin module

aspacca (approved):

PR #281, opened 2023-03-08
action: notify Coverage as GitHub comment (PR only)

v1v:
## What does this PR do?

Enable coverage reporting and storing artifacts so they can be useful for debugging purposes.

## Why is it important?

Visualise the coverage as we used to have with Jenkins

Originally removed for required extra permissions -> https://github.com/elastic/elastic-serverless-forwarder/pull/223/commits/168d8022435f04f9cf9d879d21fefcfbf81da8ab#r1094793556

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,9 @@ jobs:
   coverage:
     runs-on: ubuntu-latest
     timeout-minutes: 60
+    permissions:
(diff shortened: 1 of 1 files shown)


v1v:
There is a hard requirement to use a workflow_run since forked PRs wont' have access to the write access, see https://github.com/marocchino/sticky-pull-request-comment/issues/913

and I don't wanna enabling `write` access for now.

PR #294, opened 2023-03-16
action: notify Coverage as GitHub comment (PR only)

aspacca:
Iterating on #281: I have no push permission on @v1v fork, so I've created a separate branch on my fork
## What does this PR do?

Enable coverage reporting and storing artifacts so they can be useful for debugging purposes.

## Why is it important?

Visualise the coverage as we used to have with Jenkins

Originally removed for required extra permissions -> https://github.com/elastic/elastic-serverless-forwarder/pull/223/commits/168d8022435f04f9cf9d879d21fefcfbf81da8ab#r1094793556

### DIFF
--- a/.github/workflows/test-reporter.yml
+++ b/.github/workflows/test-reporter.yml
@@ -11,7 +11,6 @@ on:
 jobs:
   report:
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.event == 'pull_request' }}
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
         with:
@@ -20,12 +19,17 @@ jobs:
           path: "**/elastic-serverless-forwarder-junit.xml"                  # Path to test results (inside artifact .zip)
           reporter: java-junit              # Format of test results
           output-to: step-summary           # Write summary in the PR
-      - uses: dawidd6/action-download-artifact@v2
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          workflow: ${{ github.event.workflow_run.workflow_id }}
-          commit: ${{ github.event.workflow_run.head_commit.id }}
-          name: test-results
+  coverage:
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    steps:
+      - name: Download artifacts
+        run: >
+          gh run download ${{ github.event.workflow_run.id }}
+          --name test-results
+          --repo "${GITHUB_REPOSITORY}"
+          --dir .
+        env:
+          GH_TOKEN: ${{ github.token }}
       - uses: 5monkeys/cobertura-action@v13
         with:
           path: "**/coverage.xml"
--- a/.github/workflows/test-reporter.yml
+++ b/.github/workflows/test-reporter.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   report:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
     steps:
       - name: ${{ github.event.workflow_run.event }}
         run: echo "${{ github.event.workflow_run.pull_requests[0].number }}"
@@ -22,9 +23,17 @@ jobs:
           path: "**/*.xml"                  # Path to test results (inside artifact .zip)
           reporter: java-junit              # Format of test results
           output-to: step-summary           # Write summary in the PR
+      - name: dawidd6/action-download-artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: ${{ github.event.workflow_run.workflow_id }}
+          commit: ${{ github.event.workflow_run.head_commit.id }}
+          name: test-results
       - name: 5monkeys/cobertura-action
         uses: 5monkeys/cobertura-action@v13
         with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
           skip_covered: false
           minimum_coverage: 100
           fail_below_threshold: true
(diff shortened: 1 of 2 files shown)


aspacca:
@v1v that's tricky: it seems the `test-reporter` workflow is run from main, so I cannot exactly test the outcome of the changes in this PR.
I see the cobertura-action workflow checkouts the current head (https://github.com/5monkeys/cobertura-action/blob/master/.github/workflows/coverage.yml#L14-L17), but I'd guess that's just affecting the fact the action is run with the content of the PR (https://github.com/5monkeys/cobertura-action/blob/master/.github/workflows/coverage.yml#L27)

aspacca:
>I see the cobertura-action workflow checkouts the current head (https://github.com/5monkeys/cobertura-action/blob/master/.github/workflows/coverage.yml#L14-L17), but I'd guess that's just affecting the fact the action is run with the content of the PR (https://github.com/5monkeys/cobertura-action/blob/master/.github/workflows/coverage.yml#L27)

that proofed to be true with https://github.com/elastic/elastic-serverless-forwarder/pull/294/commits/a98f3a92441d1e6ca6b39e9012efd599c1be7130

.github/workflows/test-reporter.yml
```diff
@@ -11,10 +11,27 @@ on:
 jobs:
   report:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
```
v1v, line 14: This condition will affect the test-reporting for branches/tags, so what do you think creating a new job in charge of the `coverage` reporting?

```
jobs:
  report:
     ...
  coverage:
    if: ${{ github.event.workflow_run.event == 'pull_request' }}
    ...
```

.github/workflows/test-reporter.yml
```diff
@@ -11,10 +11,27 @@ on:
...
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
         with:
           artifact: test-results            # artifact name
           name: JUnit Tests                 # Name of the check run which will be created
-          path: "**/*.xml"                  # Path to test results (inside artifact .zip)
+          path: "**/elastic-serverless-forwarder-junit.xml"                  # Path to test results (inside artifact .zip)
           reporter: java-junit              # Format of test results
+          output-to: step-summary           # Write summary in the PR
+      - uses: dawidd6/action-download-artifact@v2
```
v1v, line 23: NIT:

What do you think of using the standard gh cli instead? Similar to https://github.com/elastic/apm-agent-php/blob/c2e643f76060ea731d54aa29b77109630e9dc497/.github/workflows/snapshoty.yml#L18-L25

Otherwise, let's use a pinned version

.github/workflows/test-reporter.yml
```diff
@@ -11,10 +11,27 @@ on:
...
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: ${{ github.event.workflow_run.workflow_id }}
+          commit: ${{ github.event.workflow_run.head_commit.id }}
+          name: test-results
+      - uses: 5monkeys/cobertura-action@v13
```
v1v, line 29: NIT: Use a pinned version if possible

aspacca: that's the pinned version https://github.com/5monkeys/cobertura-action/releases :)

.github/workflows/test-reporter.yml
aspacca: I was looking into it and found https://github.com/cli/cli/issues/5625

not sure how it works on `apm-agent-php` repo, I could find the dockerfile for docker.elastic.co/observability-ci/snapshoty :)

.github/workflows/test-reporter.yml
aspacca: done

.github/workflows/test-reporter.yml
v1v: I meant the commit sha instead, https://github.com/5monkeys/cobertura-action/commit/4157521550018abbbd29dfdf1b5963e907ea37f0 🙏

.github/workflows/test-reporter.yml
```diff
@@ -16,5 +16,26 @@ jobs:
         with:
           artifact: test-results            # artifact name
           name: JUnit Tests                 # Name of the check run which will be created
-          path: "**/*.xml"                  # Path to test results (inside artifact .zip)
+          path: "**/elastic-serverless-forwarder-junit.xml"                  # Path to test results (inside artifact .zip)
           reporter: java-junit              # Format of test results
+          output-to: step-summary           # Write summary in the PR
+  coverage:
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    steps:
```
v1v, line 24: Jobs need the `runs-on`:

```suggestion
    runs-on: ubuntu-latest
    steps:
```

.github/workflows/test-reporter.yml
aspacca: uhm. not a big deal. but why preferring the commit sha (and that specific one) instead of the tag? :)

v1v: They recommend to use immutable sha commits in the [using-third-party-actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) section:

> Pin actions to a full length commit SHA
Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

When possible we use that approach

PR #353, opened 2023-05-02
add SqsManagedSseEnabled to CF

### LINKED ISSUE #305: Handling of encrypted content to be sent to the Continuing and Replay queues
**Describe the enhancement:**

The content of the original inputs of the Forwarder could be encrypted: the decryption happens transparently in the Forwarder, just providing it the ARNs of the KMS Keys at deployment time so that the Forwarder lambda can be granted the proper IAM permissions required.
There can be the case where (part of) the original content is sent to the continuing queue or the replay queue: in this case the content is sent decrypted, lowering the security level of the original content.
We should provide a way to send the original content to the two above queues encrypted: my suggestion is to make the two queues encrypted by default, without the option of opting-out from encryption. We'll generate the KMS Keys at deployment time and configure the proper settings for the two queues. The two queues in the existing deployments will change to be encrypted after the given release.

Existing messages in the queues before making them encrypted will remain unencrypted: it has to be investigated if this requires special handling in the Forwarder code, or is it a blocker at all (can the lambda receive a mix of unencrypted and encrypted messages from the same queue?)

aspacca:
Enhancement
## What does this PR do?
Force Server Side Encryption on the replay and continuing queues for stack created before September/October 2022

## Why is it important?
Security concerns

## Checklist

## Author's Checklist

- [ ]

## Related issues

-
Closes #305

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -486,7 +486,7 @@ def kinesis_record_id(event_payload: dict[str, Any]) -> str:
     hex_src = get_hex_prefix(src)
 
     return f"{approximate_arrival_timestamp}-{hex_src}-{offset:012d}"
-
+    from elasticapm.metrics.base_metrics import
 
 # This is implementation specific to AWS and should not reside on share
 def expand_event_list_from_field_resolver(integration_scope: str, field_to_expand_event_list_from: str) -> str:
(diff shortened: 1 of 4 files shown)


handlers/aws/utils.py
```diff
@@ -486,7 +486,7 @@ def kinesis_record_id(event_payload: dict[str, Any]) -> str:
     hex_src = get_hex_prefix(src)
 
     return f"{approximate_arrival_timestamp}-{hex_src}-{offset:012d}"
-
+    from elasticapm.metrics.base_metrics import
```
girodav, line 489: This looks like a leftover.

girodav (commented):
Do you think we should also mention it in the docs? It's not a breaking change per se, but any update will also trigger an update of the queues.

PR #487, opened 2023-09-19
Add SAM and docker requirement into dist.sh script

kaiyan-sheng:
## What does this PR do?

This PR is adding AWS SAM CLI and docker requirement into dist.sh script.


girodav (approved):
Thanks!

PR #507, opened 2023-10-27
Handle  id for logstash

aspacca:
Enhancement
## What does this PR do?
Move `_id_` field to `@metadata._id` for logstash output

## Why is it important?
`_id` is a relevant field for elasticsearch but requires custom handling in logstash, like a rename filter.
it is a common practice (see beats behavior) to rename `_id` field to `@metadata._id` before sending the event to logstash

## Checklist

~- [] I have made corresponding changes to the documentation~
~- [] I have made corresponding change to the default configuration files~

## Author's Checklist

- [ ]

## Related issues

-


girodav (approved):
Thanks for this!

We should update the plugin docs with this change as well https://github.com/logstash-plugins/logstash-input-elastic_serverless_forwarder/blob/main/docs/index.asciidoc

Issue #181, opened 2022-11-05
Make Elastic Serverless Forwarder available in AWS GovCloud regions


ravikesarwani:
Currently, Elastic Serverless Forwarder(ESF) is not available in AWS GovCloud regions.

From the [AWS docs](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-sar.html) "Applications that are publicly shared in other AWS Regions are not automatically available in AWS GovCloud (US) Regions. To make applications available in AWS GovCloud (US) Regions, you must publish and share them independently of other AWS Regions".

This issue is to publish ESF in [AWS SAR](https://aws.amazon.com/serverless/serverlessrepo/) to be made available on the AWS GovCloud(US) Regions. 

There are certain requirements around who can have access to AWS GovCloud(US) Regions. For [example](https://aws.amazon.com/govcloud-us/?whats-new-ess.sort-by=item.additionalFields.postDateTime&whats-new-ess.sort-order=desc):
"AWS GovCloud (US-East) and (US-West) Regions are operated by employees who are U.S. citizens on U.S. soil. AWS GovCloud (US) is only accessible to U.S. entities and root account holders who pass a screening process. **Customers must confirm that they will only use a U.S. person (green card holder or citizen as defined by the U.S. Department of State) to manage and access root account keys to these regions.**"

- [ ] Figure out the logistics of AWS account that can be used for this
- [ ] Figure out the logistics of who can get access and get access to those folks
- [ ] Figure out the process of publishing ESF from that account to AWS GovCloud (US-East) and (US-West) Regions
- [ ] Define the process of how we will maintain this for each updated version over time

```[tasklist]
### Related
- [ ] https://github.com/elastic/enhancements/issues/17413
- [ ] https://github.com/elastic/enhancements/issues/18231
```

michaelmagyar:
Adding some thoughts as Functionbeat is deprecated with support ending in less than a year, and federal clients and their service providers will need extensive time to plan, implement, re-document, and possibly get re-audited on the change (i.e., they may have to add SQS, Secrets Manager, etc. to their packages if they weren't using them before).

[...]

PR #510, opened 2023-10-27
Add support for aws govcloud in dist.sh

kaiyan-sheng:
## What does this PR do?

This PR adds support for publishing elastic serverless forwarder to SAR in AWS GovCloud.
With this PR we are not adding a new config parameter. Instead, we are only checking the given region name to see if it includes `gov` in the name and set the bucket ARN, bucket resource, and a new `AWS_OR_AWS_GOV` variable.

In `macro.yaml`, we are using `%awsOrGov%` which equals to `aws` or `aws-us-gov` to construct resource/ARN. For example: if the region is not in AWS Gov, then `resource = f"arn:aws:s3:::{bucket_name_and_object_key}"` and if region is in AWS Gov, then `resource = f"arn:aws-us-gov:s3:::{bucket_name_and_object_key}"`.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-serverless-forwarder/issues/181


.internal/aws/scripts/dist.sh
```diff
@@ -67,8 +78,8 @@ find {handlers,share,shippers,storage} -not -name "*__pycache__*" -name "*.py" -
 cp -v LICENSE.txt "${CODE_URI}/LICENSE.txt"
 cp -v docs/README-AWS.md "${CODE_URI}/README.md"
 
-sed -e "s|%codeUri%|${CODE_URI}|g" -e "s/%sarAppName%/${SAR_APP_NAME}/g" -e "s/%sarAuthorName%/${SAR_AUTHOR_NAME}/g" -e "s/%semanticVersion%/${SEMANTIC_VERSION}/g" -e "s/%awsRegion%/${REGION}/g" .internal/aws/cloudformation/macro.yaml > "${TMPDIR}/macro.yaml"
-sed -e "s|%codeUri%|${CODE_URI}|g" -e "s/%sarAppName%/${SAR_APP_NAME}/g" -e "s/%sarAuthorName%/${SAR_AUTHOR_NAME}/g" -e "s/%semanticVersion%/${SEMANTIC_VERSION}/g" -e "s/%awsRegion%/${REGION}/g" -e "s/%accountID%/${ACCOUNT_ID}/g" .internal/aws/cloudformation/template.yaml > "${TMPDIR}/template.yaml"
+sed -e "s|%codeUri%|${CODE_URI}|g" -e "s/%sarAppName%/${SAR_APP_NAME}/g" -e "s/%sarAuthorName%/${SAR_AUTHOR_NAME}/g" -e "s/%semanticVersion%/${SEMANTIC_VERSION}/g" -e "s/%awsRegion%/${REGION}/g" -e "s/%awsOrGov%/${AWS_OR_AWS_GOV}/g" .internal/aws/cloudformation/macro.yaml > "${TMPDIR}/macro.yaml"
```
aspacca, line 81: since we are already using `sed` here, do we really need to add the `MacroESFAWSGov`? is not enough to replace directly `aws-us-gov`/`aws` in the template with sed?

aspacca (approved):

publish_lambda.sh
```diff
@@ -322,7 +333,7 @@ def create_policy(publish_config: dict[str, Any]):
         assert isinstance(publish_config["s3-config-file"], str)
 
         bucket_name_and_object_key = publish_config["s3-config-file"].replace("s3://", "")
-        resource = f"arn:aws:s3:::{bucket_name_and_object_key}"
+        resource = f"arn:%awsOrGov%:s3:::{bucket_name_and_object_key}"
```
aspacca, line 336: this content ends up in `publish.py`, we don't execute `sed` on this content.
If I'm not wrong, you should use `${AWS_OR_AWS_GOV}` instead.
Please test :)

publish_lambda.sh
```diff
@@ -471,18 +482,18 @@ if __name__ == "__main__":
             },
             "RoleName": "${CUSTOM_ROLE_PREFIX}ApplicationElasticServerlessForwarderRole",
             "ManagedPolicyArns": [
-                "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole",
-                "arn:aws:iam::aws:policy/service-role/AWSLambdaSQSQueueExecutionRole"
+                "arn:%awsOrGov%:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole",
```
aspacca, line 485: same as above

publish_lambda.sh
```diff
@@ -471,18 +482,18 @@ if __name__ == "__main__":
...
+                "arn:%awsOrGov%:iam::aws:policy/service-role/AWSLambdaSQSQueueExecutionRole"
```
aspacca, line 486: same as above

publish_lambda.sh
```diff
@@ -471,18 +482,18 @@ if __name__ == "__main__":
...
             ],
         },
     }
 
     if vpc_config:
-        customRole["Properties"]["ManagedPolicyArns"].append("arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole")
+        customRole["Properties"]["ManagedPolicyArns"].append("arn:%awsOrGov%:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole")
```
aspacca, line 492: same as above

publish_lambda.sh
```diff
@@ -471,18 +482,18 @@ if __name__ == "__main__":
...
 
     has_kinesis_events: bool = len([created_event for created_event in created_events if created_events[created_event]["Type"] == "Kinesis"]) > 0
     if has_kinesis_events:
-        customRole["Properties"]["ManagedPolicyArns"].append("arn:aws:iam::aws:policy/service-role/AWSLambdaKinesisExecutionRole")
+        customRole["Properties"]["ManagedPolicyArns"].append("arn:%awsOrGov%:iam::aws:policy/service-role/AWSLambdaKinesisExecutionRole")
```
aspacca, line 496: same as above

aspacca (changes requested):

publish_lambda.sh
kaiyan-sheng: Ahh I didn't realize this is going into publish.py! Thanks!! I just made the changes and tested them.

aspacca (approved):

Issue #472, opened 2023-08-24
Use Localstack >= 2.x


girodav:
Localstack 2.0 introduced a new, [more performant Lambda provider](https://localstack.cloud/blog/2023-03-29-announcing-localstack-2.0-general-availability/). Since localstack is used heavily in our integration tests, we should consider upgrading from 1.4.0. 

_Note: in https://github.com/elastic/elastic-serverless-forwarder/pull/399, we attempted to upgrade to localstack v2.1.0, but our integration tests were failing. We should understand why and fix accordingly_

PR #534, opened 2023-11-23
Remove delete sqs message calls and optimise json collector decorator

aspacca:
Enhancement

## What does this PR do?

Remove calls to `sqs.DeleteMessage` that are the most expensive operation in the forwarder, and optimise `JsonCollector` decorator in order to parse as json the storage content only when strictly required

It also bumps the version of Localstack used to 3.x
## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/472

### DIFF
--- a/share/multiline.py
+++ b/share/multiline.py
@@ -20,8 +20,15 @@ default_multiline_timeout: int = 5  # Default timeout in secs to finish a multi-
 
 timedelta_circuit_breaker: datetime.timedelta = datetime.timedelta(seconds=5)
 
+# CollectTuple is a tuple representing the multilines bytes content, the length of the content and the newline found
+# These data is instrumental to the `StorageDecoratorIterator` that needs the content to yield, the starting and ending
+# offsets and the newline
 CollectTuple: TypeAlias = tuple[bytes, int, bytes]
+
+# CollectIterator yields a `CollectTuple`
 CollectIterator: TypeAlias = Iterator[CollectTuple]
+
+# FeedIterator yields a tuple representing the content and its newline to feed of the `ProtocolMultiline` implementation
 FeedIterator: TypeAlias = Iterator[tuple[bytes, bytes]]
 
 
--- a/share/multiline.py
+++ b/share/multiline.py
@@ -74,7 +74,7 @@ class CollectBuffer:
         self._processed_lines: int = 0
         self._current_length: int = 0
 
-    def collect_and_reset(self) -> tuple[bytes, int]:
+    def collect_and_reset(self) -> tuple[bytes, int, bytes]:
         data = self._buffer
         current_length = self._current_length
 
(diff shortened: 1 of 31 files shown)


tests/handlers/aws/test_integrations.py
```diff
@@ -72,9 +72,10 @@ def setUpClass(cls) -> None:
         lgc = LogstashContainer(es_container=esc)
         cls.logstash = lgc.start()
 
-        lsc = LocalStackContainer(image="localstack/localstack:1.4.0")
+        lsc = LocalStackContainer(image="localstack/localstack:3.0.0")
```
girodav, line 75: Thanks for this!

We could use `3.0.1` here instead. We should have a look at how to automate these changes with dependabot.

share/multiline.py
girodav, line 1: NIT: It would be great, even as separate PR, to have a comment briefly explaining what this code does. The same applies for the overall json processing logic.

girodav (approved):

share/multiline.py
```diff
@@ -21,17 +24,17 @@ class CommonMultiline(metaclass=ABCMeta):
     Common class for Multiline components
     """
 
-    _feed: Iterator[tuple[bytes, bytes, int]]
```
tommyers-elastic, line 24: i think defining a type alias for this iterator would make this more readable (and isolate any changes like this one)

share/multiline.py
```diff
@@ -85,7 +88,14 @@ def collect_and_reset(self) -> tuple[bytes, int]:
 
         self.previous = b""
 
-        return data, current_length
+        if data.find(b"\r\n") > -1:
```
tommyers-elastic, line 91: what's the reason for this addition? i think a comment here would help future readers.

tommyers-elastic:
thanks for the PR! i'm still working through, but a couple of more general comments: 

there's lots of changes here to the JSON logic - it would be really useful to have a detailed description of what has changed and why in the PR description. 

additionally, in the future, it would be great to split the changes into seperate PRs to keep them logically seperate, and easier to review. here there are 3 changes: sqs delete message, localstack, and json parser.

tests/storage/test_benchmark.py
```diff
@@ -68,7 +68,7 @@ def get_by_lines_parameters() -> list[tuple[int, str, bytes]]:
             _IS_MULTILINE_PATTERN,
             _IS_MULTILINE_WHILE,
         ]:
-            for newline in [b"", b"\n", b"\r\n"]:
```
tommyers-elastic, line 71: i'm guessing empty `newline` denotes either the document isn't multiline format, or we have reached the end of the newline delimited data? how come we're removing the case here?

aspacca: good catch, leftover from changes during debugging, I will revert it!

share/multiline.py
aspacca: answered [here](https://github.com/elastic/elastic-serverless-forwarder/pull/534/files#diff-2e204b34a51f1f25840268e768056fd46938d1ae5320dcf0866c6092d8614c10R23-R25)

aspacca:
I will add unit test today for the decorators, then we can merge

Issue #536, opened 2023-11-23
Include ESF specific User-Agent header in outgoing Elasticsearch requests


tommyers-elastic:
This is a common pattern with [Beats](https://github.com/elastic/elastic-agent-libs/blob/main/useragent/useragent.go#L27), and will help with troubleshooting ESF issues in Elastic Cloud, as well as helping us to better understand customer usage of ESF within Elastic Cloud. 

The User-Agent header must conform to https://www.rfc-editor.org/rfc/rfc9110.html#name-user-agent, and we should be especially concious of the following:

```
A user agent SHOULD NOT generate a User-Agent header field containing needlessly fine-grained detail and SHOULD limit the addition of subproducts by third parties. Overly long and detailed User-Agent field values increase request latency and the risk of a user being identified against their wishes ("fingerprinting").
```

At a minimum, the header should contain the ESF version and the Python/Lambda runtime version.

PR #537, opened 2023-11-23
Add ESF specific User-Agent header in outgoing Elasticsearch requests

girodav:
## What does this PR do?

* Include ESF specific User-Agent header in outgoing Elasticsearch requests
* Bump the minimum supported version for the Elastic stack to 7.17

## Why is it important?

This is a common pattern with [Beats](https://github.com/elastic/elastic-agent-libs/blob/main/useragent/useragent.go#L27), and will help with troubleshooting ESF issues in Elastic Cloud, as well as helping us to better understand customer usage of ESF within Elastic Cloud.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/536

### DIFF
--- a/share/utils.py
+++ b/share/utils.py
@@ -2,16 +2,13 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 import hashlib
-import os
-
-from share.version import version
+import sys
 
 
 def get_hex_prefix(src: str) -> str:
     return hashlib.sha3_384(src.encode("utf-8")).hexdigest()
 
 
-def create_user_agent() -> str:
-    """Creates the 'User-Agent' header given the library name and version"""
-    # AWS_EXECUTION_ENV – The runtime identifier, prefixed by AWS_Lambda_ (for example, AWS_Lambda_Python3.9).
-    return f"ElasticServerlessForwarder/{version} ({os.getenv('AWS_EXECUTION_ENV')})"
+def create_user_agent(esf_version: str, environment: str = sys.version) -> str:
+    """Creates the 'User-Agent' header given ESF version and running environment"""
+    return f"ElasticServerlessForwarder/{esf_version} ({environment})"
--- a/share/utils.py
+++ b/share/utils.py
@@ -13,4 +13,5 @@ def get_hex_prefix(src: str) -> str:
 
 def create_user_agent() -> str:
     """Creates the 'User-Agent' header given the library name and version"""
-    return f"ElasticServerlessForwarder/{version} ({os.getenv('AWS_EXECUTION_ENV')}; OTHER)"
+    # AWS_EXECUTION_ENV – The runtime identifier, prefixed by AWS_Lambda_ (for example, AWS_Lambda_Python3.9).
+    return f"ElasticServerlessForwarder/{version} ({os.getenv('AWS_EXECUTION_ENV')})"
--- a/share/utils.py
+++ b/share/utils.py
@@ -2,7 +2,15 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 import hashlib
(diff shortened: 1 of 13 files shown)


share/utils.py
```diff
@@ -2,7 +2,16 @@
...
 
 
 def get_hex_prefix(src: str) -> str:
     return hashlib.sha3_384(src.encode("utf-8")).hexdigest()
+
+
+def create_user_agent() -> str:
+    """Creates the 'User-Agent' header given the library name and version"""
+    # AWS_EXECUTION_ENV – The runtime identifier, prefixed by AWS_Lambda_ (for example, AWS_Lambda_Python3.9).
+    return f"ElasticServerlessForwarder/{version} ({os.getenv('AWS_EXECUTION_ENV')})"
```
tommyers-elastic, line 17: nit: best to assume (even though this may never be the case), that we are not always running on AWS. 

we could populate the header with information available regardless of environment e.g. (`sys.version`), and conditionally add additional environment specific data where available.

shippers/es.py
```diff
@@ -119,6 +120,7 @@ def _elasticsearch_client(**es_client_kwargs: Any) -> Elasticsearch:
         es_client_kwargs["max_retries"] = 4
         es_client_kwargs["http_compress"] = True
         es_client_kwargs["retry_on_timeout"] = True
+        es_client_kwargs["headers"] = {"user-agent": share.utils.create_user_agent()}
```
tommyers-elastic, line 123: super tiny nit: i know headers are case-insensitive, but I usually see `User-Agent`.

share/utils.py
```diff
@@ -2,7 +2,16 @@
...
+import os
+
+from share.version import version
 
 
 def get_hex_prefix(src: str) -> str:
     return hashlib.sha3_384(src.encode("utf-8")).hexdigest()
+
+
+def create_user_agent() -> str:
```
tommyers-elastic, line 14: can we pass the verison in to make this more easily testable?

share/version.py
```diff
@@ -0,0 +1,5 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+version = "1.9.0"
```
aspacca, line 5: let's make the content of this file updated on deploy time or in CI when we release a new tag. no way we have to remember to change its content on every release :)

share/utils.py
aspacca: more then "nit" :)

if it's in share it should have no reference to the cloud environment :)

we could pass the cloud specific data as parameter

share/version.py
girodav: I agree, however, there are different approaches and caveats to consider:

* Should we have a workflow that is triggered on a `git tag push` event, that bumps this line to the next version? Since we do not have version branches, this wouldn't be that trivial to implement. What do we do if the current version in main is, for example, `1.11.1`, but we want to create a tag for `1.20`?

* Should we bump the version based on [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)? E.g we add [bugfix] as prefix to the commit, and this automatically bump the minor version. How do we automatically handle multiple features/commit in a single release?

* We do not provide support for other cloud providers apart from AWS, yet the version file is in the shared package. I would not worry about it right now, as no support for other cloud providers is in the horizon, but I'd keep in mind while choosing the best approach. Related to this, the current tag name pattern does not match what it is written in `version.py` (i.e lambda-vX.X.X instead of vX.X.X)

tommyers-elastic (approved):
LGTM

tommyers-elastic (approved):

share/version.py
girodav: I agree. I moved the discussion on this tracking issue https://github.com/elastic/elastic-serverless-forwarder/issues/540

Issue #547, opened 2023-12-02
Add ESF lambda ARN to SAR output


vfeinman:
**Describe the enhancement:**
Bubble up the ESF lambda ARN through stack/application output so that any ESF lambda dependent resources can be created based on the lambda ARN generated by the application. 

**Describe a specific use case for the enhancement or feature:**
On a number of occasions the ESF lambda concurrent executions spiked, consuming the account limit for Lambda concurrent executions. This has the undesirable effect of preventing production applications from running and extremely high costs. We currently have to deploy the ESF Serverless Application Repository application, then manually set the reserved concurrent executions manually. I'd also like the ability to have native AWS CloudWatch monitoring the ESF lambda, which I'm not able to do in a scalable way since I don't have any access to lambda ARN created through the ESF application.

PR #552, opened 2023-12-07
Add SAM Template outputs for Lambda function ARN and IAM Role ARN

girodav:
## What does this PR do?

Add the following Outputs to the SAM template:
* ESF Lambda function ARN
* ESF Lambda function IAM role ARN

The PR also fixes a bug within `dist.sh` that prevented non-gov AWS regions from being recognized correctly.

## Why is it important?

It allows users to create dependent resources based on the Lambda Function ARN.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/547

### DIFF
--- a/.internal/aws/cloudformation/application.yaml
+++ b/.internal/aws/cloudformation/application.yaml
@@ -60,6 +60,7 @@ Resources:
   ApplicationElasticServerlessForwarder:
     Type: AWS::Serverless::Function
     Properties:
+      FunctionName: %sarAppName%-lambda-function
       Timeout: 900
       MemorySize: 512
       CodeUri: %codeUri%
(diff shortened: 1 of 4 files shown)


.internal/aws/cloudformation/application.yaml
```diff
@@ -60,6 +60,7 @@ Resources:
   ApplicationElasticServerlessForwarder:
     Type: AWS::Serverless::Function
     Properties:
+      FunctionName: %sarAppName%-lambda-function
```
aspacca, line 63: will this trigger any consequence when updating a lambda with the previous version of the cf template?

aspacca (approved):

.internal/aws/cloudformation/application.yaml
girodav: That was one of my doubts. `FunctionName` is passed directly to the [FunctionName](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-functionname) property of an AWS::Lambda::Function resource. This requires replacing the function so updating to this version will cause downtime, which will occur anyway since the underlying code signature is different. I think also LogGroups and Metric Names in Cloudwatch would be different
I am not aware of anything else, unless we use the `FunctionName` elsewhere. 

We might remove it since the AWS Lambda Function name can be extracted from the ARN. The impact sounds bigger than what I originally anticipated. What do you think?

aspacca: yes, I'd remove

aspacca (approved):

PR #544, opened 2023-12-01
Remove delete sqs message calls and optimise json collector decorator

aspacca:
Enhancement

## What does this PR do?

Remove calls to `sqs.DeleteMessage` that are the most expensive operation in the forwarder, and optimise `JsonCollector` decorator in order to parse as json the storage content only when strictly required

It also bumps the version of Localstack used to 3.x
## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have made corresponding change to the default configuration files~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `CHANGELOG.md`

## Author's Checklist

- [ ]

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/472

PR #553, opened 2023-12-08
Fix json collector decorator and switch json parsing library to orjson

aspacca:
Bug/Enhancement
## What does this PR do?
Fix a bug related to event list from field expander where, if json content type was single and spanning more than 1000 lines the circuit breaker of the json decorator will kick in

It also switch orjson parsing library over ujson one

## Why is it important?
A regression was introduced on json decorator while refactored with #544 
Also switching from usjon to orjson improves performances

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -313,8 +313,11 @@ def get_trigger_type_and_config_source(event: dict[str, Any]) -> tuple[str, str]
                 and "eventSource" in body["Records"][0]
             ):
                 event_source = body["Records"][0]["eventSource"]
+                if event_source not in _available_triggers:
+                    raise Exception
             else:
                 raise Exception
+
         except Exception:
             if "eventSource" not in first_record:
                 raise Exception("Not supported trigger")
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -314,9 +314,9 @@ def get_trigger_type_and_config_source(event: dict[str, Any]) -> tuple[str, str]
             ):
                 event_source = body["Records"][0]["eventSource"]
                 if event_source not in _available_triggers:
-                    raise Exception
+                    raise Exception("except in the function")
             else:
-                raise Exception
+                raise Exception("except in the function")
 
         except Exception:
             if "eventSource" not in first_record:
--- a/storage/decorator.py
+++ b/storage/decorator.py
@@ -236,6 +236,21 @@ def json_collector(
 
             yield line, _, _, newline, None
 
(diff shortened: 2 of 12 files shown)


handlers/aws/utils.py
```diff
@@ -313,8 +313,11 @@ def get_trigger_type_and_config_source(event: dict[str, Any]) -> tuple[str, str]
                 and "eventSource" in body["Records"][0]
             ):
                 event_source = body["Records"][0]["eventSource"]
+                if event_source not in _available_triggers:
+                    raise Exception
```
girodav, line 317: I would add an exception message of some kind here

storage/decorator.py
```diff
@@ -236,6 +236,21 @@ def wrapper(
 
             yield line, _, _, newline, None
 
+    def _collect_single(iterator: StorageDecoratorIterator) -> StorageDecoratorIterator:
+        single: list[tuple[Union[StorageReader, bytes], int, int, bytes]] = list(
+            [
+                (data, starting_offset, ending_offset, newline)
+                for data, starting_offset, ending_offset, newline, _ in iterator
+            ]
+        )
```
girodav, line 245: I find this DS a bit hard to digest. Maybe a comment with an example entry could help

storage/decorator.py
```diff
@@ -255,43 +270,36 @@ def wrapper(
...
+            if storage.json_content_type == "single":
+                iterator = _collect_single(iterator=iterator)
+                json_collector_state.is_a_json_object = True
+                json_collector_state.has_an_object_start = True
+
+            for data, starting_offset, ending_offset, newline, _ in iterator:
+                assert isinstance(data, bytes)
+
+                # if it's not a json object we can just forward the content by lines
+                if not json_collector_state.has_an_object_start:
```
girodav, line 284: This comment does not seem to be consistent with the if statement below. Should this use `json_collector_state.is_a_json_object` instead?

girodav (commented):
Thanks for this. I wonder if we should also make the PR title more descriptive. Pre-release sounds a bit generic.

handlers/aws/utils.py
aspacca: I added a message only where the exceptions will bubble up to the function callar

aspacca:
@girodav addressed your comment, please review :)

girodav (approved):

aspacca:
@girodav , I fixed a couple of comments and review status was dismissed

feel free to re-approve and merge, thanks :)

girodav (approved):

PR #627, opened 2024-02-20
Beyond 4096b cf param sar

### LINKED ISSUE #560: Increasing cloudwatch log groups subscription limit
Hi, I am exploring the serverless forwarder to forward the logs of lambda from aws account to Elasticsearch. Upon deploying the stack, I found out that it hits the CloudFormation quota limit. To be specific, the parameter value for `ElasticServerlessForwarderCloudWatchLogsEvents` cannot exceed 4096 bytes. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

I tried several ways to workaround this hard limit, namely joining multiple variables to form `ElasticServerlessForwarderCloudWatchLogsEvents` in CloudFormation Stacks, or deploying multiple forwarders in the same accounts. However both approaches fail. The 1st approach fail due to nested stack containing the single `ElasticServerlessForwarderCloudWatchLogsEvents` parameter, and the 2nd approach fail due to name clash in the marco.

Will it be possible to support this needs with a new version? Or any other workarounds possible (deploying marco and forwarder separately or using AWS integration instead)? In the future, we would like to send logs from hundreds of lambdas to Elastic Cloud, so this ceiling is blocking us. Thank you for your help.

aspacca:
Enhancement
## What does this PR do?
Add a second CF parameter for event triggers in the SAR deployment

## Why is it important?
There is a limit of 4096b on the length of CF parameter, adding an extra second one to mitigate the limit

## Checklist

## Author's Checklist

- [ ]

## Related issues

-
Closes #560

### DIFF
--- a/.internal/aws/cloudformation/macro.yaml
+++ b/.internal/aws/cloudformation/macro.yaml
@@ -38,8 +38,42 @@ Resources:
                         }
                     }
 
+            if "ElasticServerlessForwarderKinesisEvents2" in parameters:
+                for kinesis_event in parameters["ElasticServerlessForwarderKinesisEvents2"]:
+                    kinesis_event = kinesis_event.strip()
+                    if len(kinesis_event) == 0:
+                        continue
+
+                    kinesis_event_name = f"KinesisEvent{hex_suffix(kinesis_event)}"
+                    events_fragment[kinesis_event_name] = {
+                        "Type": "Kinesis",
+                        "Properties": {
+                            "Stream": kinesis_event,
+                            "StartingPosition": "TRIM_HORIZON",
+                            "BatchSize": 10,
+                            "Enabled": True,
+                        }
+                    }
+
             if "ElasticServerlessForwarderSQSEvents" in parameters:
                 for sqs_event in parameters["ElasticServerlessForwarderSQSEvents"]:
+                    sqs_event = sqs_event.strip()
+                    if len(sqs_event) == 0:
+                        continue
+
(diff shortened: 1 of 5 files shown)


.internal/aws/cloudformation/macro.yaml
```diff
@@ -38,6 +38,23 @@ Resources:
                         }
                     }
 
+            if "ElasticServerlessForwarderKinesisEvents2" in parameters:
```
kaiyan-sheng, line 41: Overall it looks good to me. Not sure if it's worth extracting all the sub functions like this:
```
def process_sqs_events(events, events_fragment):
    for sqs_event in events:
        sqs_event = sqs_event.strip()
        if len(sqs_event) == 0:
            continue
        sqs_event_name = f"SQSEvent{hex_suffix(sqs_event)}"
        events_fragment[sqs_event_name] = {
            "Type": "SQS",
            "Properties": {
                "Queue": sqs_event,
                "BatchSize": 10,
                "Enabled": True,
            }
        }

if "ElasticServerlessForwarderSQSEvents" in parameters:
    process_sqs_events(parameters["ElasticServerlessForwarderSQSEvents"], events_fragment)

if "ElasticServerlessForwarderSQSEvents2" in parameters:
    process_sqs_events(parameters["ElasticServerlessForwarderSQSEvents2"], events_fragment)
```
Not a required change, either way works!

kaiyan-sheng (approved):

PR #665, opened 2024-03-22
docs: update link to documentation

Comradin:
The former page is outdated and mentions to go to a different page
instead.

Update the original link reference to the correct page and skip the
intermediate one.

fixes 664


Comradin:
Signed the Contributor Agreement .. Just a bit puzzled about that title field. 😁

Comradin:
Signed with work address and github handle now

florianl (approved):
Thanks for your contribution!

Issue #683, opened 2024-04-11
Create pipeline to push zip file with dependencies to an S3 bucket


constanca-m:
## Description

This issue comes from [this](https://github.com/elastic/terraform-elastic-esf/pull/1#discussion_r1516188875) comment thread of a PR to use terraform to install ESF.

The current approach for the terraform files:
- Download this repository files based on the release version
- Use this handler for the lambda function handler by using a docker environment and installing all dependencies

The desired approach: have all dependencies in a zip file and push this to an S3 bucket.

## Steps

**Step 1**

Create a new buildkite pipeline in this directory.

Each version release (or commit?) triggers the creation of a new zip file with all the dependencies. This zip file needs to be pushed to an S3 bucket that will be used by customers. The S3 bucket needs to be read only. 

The zip file will have the following structure:

- It will have a directory for each needed package mentioned in [requirements.txt](https://github.com/elastic/elastic-serverless-forwarder/blob/main/requirements.txt). Package is generated through the use of command `pip install --target ./package <REQUIREMENT>`.
- It will have the [file](https://github.com/elastic/elastic-serverless-forwarder/blob/main/main_aws.py) with the code for the handler at the root. Because this file points to others, we will push the [handlers directory](https://github.com/elastic/elastic-serverless-forwarder/tree/main/handlers) as well.

Reference: https://docs.aws.amazon.com/lambda/latest/dg/python-package.html#python-package-create-dependencies.

**Step 2**
Refactor the terraform files:
- Remove the need to [download this repository.](https://github.com/elastic/terraform-elastic-esf/blob/80c8f7418fcfe01bd6516c119bb61cb656efbdfc/esf.tf#L60-L69)
- Remove the [module](https://github.com/elastic/terraform-elastic-esf/blob/80c8f7418fcfe01bd6516c119bb61cb656efbdfc/esf.tf#L73-L134) for lambda.
- Insert a new resource `aws_lambda_function` that reads from the S3 bucket with the zip file:

> 
> ```tf
> resource "aws_lambda_function" "esf" {
> // stuff
> 
>   s3_bucket = aws_s3_bucket.esf_bucket.id
>   s3_key    = aws_s3_object.esf_zip_bundle.key
> 
> // stuff
> }
> ```
> _Originally posted by @girodav in https://github.com/elastic/terraform-elastic-esf/pull/1#discussion_r1516188875_

## Tasks

[...]

Issue #540, opened 2023-11-28
Version Bump Automation


girodav:
This issue tracks the discussion around handling major/minor version bump automatically.

- Should we have a workflow that is triggered on a git tag push event, that bumps this line to the next version? Since we do not have version branches, this wouldn't be that trivial to implement. What do we do if the current version in main is, for example, 1.11.1, but we want to create a tag for 1.20?

 - Should we bump the version based on [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)? E.g we add [bugfix] as prefix to the commit, and this automatically bump the minor version. How do we automatically handle multiple features/commit in a single release?

 - We do not provide support for other cloud providers apart from AWS, yet the version file is in the shared package. I would not worry about it right now, as no support for other cloud providers is in the horizon, but I'd keep in mind while choosing the best approach. Related to this, the current tag name pattern does not match what it is written in version.py (i.e `lambda-vX.X.X` instead of `vX.X.X`)

One possible approach is to trigger a github action to "release" (i.e creating git tag and possibly deploy on SAR) whenever `version.py` is modified. 

```[tasklist]
### Related tasks
- [ ] https://github.com/elastic/elastic-serverless-forwarder/issues/279
```

tommyers-elastic:
thanks for opening the issue @girodav 

my preference would be to track the version number using git tags as the primary source. we can create a version.py file at deploy time (via the script), which can be used at runtime to populate the version number. if we try to deploy a version that already exists, the deploy script should fail - how can we track this part? is there a way to query the existing version in SAR? 

we also have to remove the manual tracking of the version number in the changelog file. one way to do this would be to utillize something like the [elastic-agent-changelog-tool](https://github.com/elastic/elastic-agent-changelog-tool), but we could just as easily roll some simple automation triggered off new git tags, which does something like this and then commits the new file before writing the tag. 

```
changelog="./CHANGELOG.md"
version=$(git describe --tags --match="v[0-9]*" --abbrev=0)
today=$(date -u '+%Y\/%m\/%d')
sed -i -e "s/<version.next>/$version - $today/g" $changelog
echo "### <version.next>\n\n$(cat $changelog)" > $changelog
```

PR #685, opened 2024-04-15
[Github workflow] Add new tag on version.py update

constanca-m:
## What does this PR do?

It adds two GitHub actions workflows:
- `changelog`: when `version.py` is updated on a pull request, it will check if `CHANGELOG` has the first line with `* v$VERSION*`
- `release`: When `share/version.py` gets pushed, it will automatically create a new tag named `lambda-v{VERSION}`.

## Why is it important?

It automates the tag process as mentioned in https://github.com/elastic/elastic-serverless-forwarder/issues/540.

With this workflow, we can also create a new workflow to zip dependencies on a new tag creation like mentioned in the issue https://github.com/elastic/elastic-serverless-forwarder/issues/683.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-serverless-forwarder/issues/540.
- Relates https://github.com/elastic/elastic-serverless-forwarder/issues/683.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,37 @@
+---
+## Workflow to check CHANGELOG was updated when version gets updated in share/version.py
+name: changelog
+
+on:
+  pull_request:
+    paths:
+      - 'share/version.py'
+
+jobs:
+
+  changelog:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+
+      - uses: actions/checkout@v4
+
+      - name: Get version number
+        shell: bash
+        run: |
+          VERSION=$(awk '/version/ {print $3}' share/version.py | tr -d '"')
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+          echo "::notice::ESF version is $VERSION."
+
+      - name: Check changelog
+        shell: bash
+        run: |
+          first_line=$(awk 'NR==1' CHANGELOG.md)
+
+          if [[ $first_line != *" v${{ env.VERSION }} "* ]]; then
+            error="CHANGELOG should contain the new version in its first line, like this ' v1.1.1 '."
+            reminder="Do not forget to add space before and after the version in the CHANGELOG first line."
+            echo "::error::$error $reminder"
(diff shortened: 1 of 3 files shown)


girodav:
Have you considered a different approach, like updating `version.py` when a git tag is created instead? (see Tom's comment [here](https://github.com/elastic/elastic-serverless-forwarder/issues/540#issuecomment-1830305924)). 

Regardless of the approach, I'd also make sure that also CHANGELOG.md is updated before releasing.

constanca-m:
Thanks @girodav. I added a changelog workflow. I will leave the version.py as the trigger, as it does not seem to make much difference.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,37 @@
...
+    timeout-minutes: 5
+
+    steps:
+
+      - uses: actions/checkout@v4
+
+      - name: Get version number
+        shell: bash
+        run: |
+          VERSION=$(awk '/version/ {print $3}' share/version.py | tr -d '"')
```
gizas, line 23: Consider 
shell grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' /share/version.py

https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/Makefile#L5C16-L5C74

(I am afraid of someone putting an extra space in file and breaks the automation)

constanca-m: Can you explain this part 

> I am afraid of someone putting an extra space in file and breaks the automation

a bit more @gizas? Maybe by an example?  I don't understand why it would break

gizas: version = " 1.13.1" 

then

```bash
awk '/version/ {print $4}' version.py | tr -d '"'  ---> Thins needs to become $4
1.13.1
❯ grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' version.py ---> this still works
1.13.1
```

constanca-m: Thank you! I checked and you are corrected. I updated it.

PR #687, opened 2024-04-17
Update Logstash/ES versions being used in the tests

constanca-m:
The newest PRs all fail the `coverage` workflow.

This happens because `test_integrations` needs to create containers for the tests. The container logstash was getting stuck in `waiting to be ready...`.

Updating the ES and Logstash versions to 7.17.20 solves this issue.


tests/scripts/run_tests.sh
```diff
@@ -8,7 +8,7 @@ set -ex
 # delete any __pycache__ folders to avoid hard-to-debug caching issues
 find . -name __pycache__ -type d -exec rm -r '{}' + || true
 PYTEST_ARGS=("${PYTEST_ARGS}")
-py.test -vv "${PYTEST_ARGS[*]}" "${PYTEST_JUNIT}" tests
+py.test -vv "${PYTEST_ARGS[*]}" "${PYTEST_JUNIT}" tests -s
```
constanca-m, line 11: I am leaving the `-s` flag here so we can see the logs when `make coverage` runs. Otherwise, if the tests get stuck again, we have no way to know what is stopping them.

gizas (approved):

PR #689, opened 2024-04-17
Upload dependencies to S3 bucket on new tag release

constanca-m:
## What does this PR do?

Uploads dependencies to S3 bucket on new tag release.

This workflow is dependent on workflow `release` - that is, it waits for the new release to zip all the dependencies and send it to S3 bucket.

You can see more details discussed in issue https://github.com/elastic/elastic-serverless-forwarder/issues/683.

## Checklist

## Related issues

Closes https://github.com/elastic/elastic-serverless-forwarder/issues/683.

### DIFF
--- a/.github/workflows/upload-dependencies.yml
+++ b/.github/workflows/upload-dependencies.yml
@@ -1,0 +1,66 @@
+---
+## Workflow to push zip with dependencies to S3 bucket every time the ESF version is updated
+name: upload-dependencies
+
+env:
+  BUCKET_NAME : "esf-dependencies"
+  AWS_REGION : "eu-central-1"
+  ROLE: "arn:aws:iam::267093732750:role/esf-dependencies-role"
+
+
+on:
+  workflow_run:
+    workflows: [release]
+    types:
+      - completed
+
+
+permissions:
+  id-token: write   # This is required for requesting the JWT
+  contents: read    # This is required for actions/checkout
+
+
+jobs:
+
+  build-and-upload-dependencies:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      # See https://docs.aws.amazon.com/lambda/latest/dg/python-package.html#python-package-create-dependencies
+
+      - uses: actions/checkout@v4
+
+      - name: Get version number
+        shell: bash
+        run: |
+          VERSION=$(awk '/version/ {print $3}' version.py | tr -d '"')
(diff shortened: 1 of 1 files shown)


.github/workflows/upload-dependencies.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+      # See https://docs.aws.amazon.com/lambda/latest/dg/python-package.html#python-package-create-dependencies
+
+      - uses: actions/checkout@v4
+
+      - name: Get version number
+        shell: bash
+        run: |
+          VERSION=$(grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' share/version.py)
```
axw, line 37: Perhaps get the tag from git, to avoid duplicating this logic across multiple workflows? e.g. we could use `git describe` to get the tag for the current branch.

axw (commented):
The changes in this PR look good - nice work.

https://github.com/elastic/elastic-serverless-forwarder/pull/685 makes me a little nervous - IIUC we could recreate the tag whenever a change is made to version.py, which seems a little too implicit for my liking.

I think we should make sure some tests are run before releasing any new artifacts. That should probably be done before tagging.

constanca-m:
> IIUC we could recreate the tag whenever a change is made to version.py, which seems a little too implicit for my liking.

Indeed, I noticed it too, but I figured there is really no reason to change that file unless to update the version.

> I think we should make sure some tests are run before releasing any new artifacts.

What kind of tests? There is the test workflow (`make coverage`) that is running all the tests, any missing?

I don't mind recreating the PR to make a few changes @axw

axw:
> What kind of tests? There is the test workflow (make coverage) that is running all the tests, any missing?

Ah I see. Is that guaranteed to have run before the release workflow?

I guess that's OK to start with, but ideally we would run some smoke tests too, preferably before even tagging. But anyway that can come in a followup.

For reference, apm-aws-lambda runs smoke tests on every push to main: https://github.com/elastic/apm-aws-lambda/blob/main/.github/workflows/smoke-tests.yml. Maybe a little excessive, but there might be some workflows and code that could be reused.

constanca-m:
> Ah I see. Is that guaranteed to have run before the release workflow?

Yes, it runs on every PR. The release only works on a push, so only after the tests.

axw (approved):

PR #690, opened 2024-04-18
Fix tag release workflow

constanca-m:
## What does this PR do?

The current release workflow was failing with:

```
Run actions/github-script@v7
SyntaxError: Unexpected number
    at new AsyncFunction (<anonymous>)
Error: Unhandled error: SyntaxError: Unexpected number
    at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v7/dist/index.js:35424:16)
    at main (/home/runner/work/_actions/actions/github-script/v7/dist/index.js:35522:26)
    at /home/runner/work/_actions/actions/github-script/v7/dist/index.js:35497:1
    at /home/runner/work/_actions/actions/github-script/v7/dist/index.js:35553:3
    at Object.<anonymous> (/home/runner/work/_actions/actions/github-script/v7/dist/index.js:35556:12)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
```

Putting the version variable inside `"` to make it a string solves this issue.

We also check if the version variable in `version.py` increased. If that is the case, then we create a new release. Otherwise, nothing happens.

## Checklist

## How to test

As far as I am aware, the only way to test would be by using your own repository. Just do the following:

- Have a `share/version.py` file there.
- Copy the release workflow to that repository.

Then make the commits to check everything is working as expected. You need to update the `version.py` file to trigger the workflow, unless you also change the trigger or make the workflow run manually.

It is not possible to test it here, because we would have to update the `version.py` file of ESF, which does not make sense in this case.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,7 @@ jobs:
           do
             if [[ ${new_numbers[i]} > ${old_numbers[i]} ]]
             then
-              create_tag=true
+              CREATE_TAG==true
               break
             elif [[ ${new_numbers[i]} < ${old_numbers[i]} ]]
             then
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,7 +59,7 @@ jobs:
           echo "CREATE_TAG=${CREATE_TAG}" >> $GITHUB_ENV
           echo "::notice::Latest version is $old_version."
           echo "::notice::Current version is ${{ env.VERSION }}."
-          echo "::notice::The result for creating tag is $create_tag."
+          echo "::notice::The result for creating tag is $CREATE_TAG."
 
       - name: Create tag
         if: env.CREATE_TAG == 'true' # run only in case CREATE_TAG is true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,22 +20,55 @@ jobs:
 
     steps:
 
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
-    - name: Get version number
-      shell: bash
-      run: |
-        VERSION=$(grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' share/version.py)
-        echo "VERSION=${VERSION}" >> $GITHUB_ENV
-        echo "::notice::ESF version is $VERSION."
+      - name: Get version number
+        shell: bash
+        run: |
+          VERSION=$(grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' share/version.py)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+          echo "::notice::ESF version is $VERSION."
 
-    - name: Create tag
-      uses: actions/github-script@v7
-      with:
-        script: |
-          github.rest.git.createRef({
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            ref: 'refs/tags/lambda-v' + "${{ env.VERSION }}",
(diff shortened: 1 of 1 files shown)


gizas (approved):

axw (dismissed):
LGTM. I'm wondering if we should make the tagging a `workflow_dispatch` workflow that one needs to explicitly run. Then we could update the version ahead of the release, and it would be present in pre-release builds and smoke tests. Doesn't need to be addressed in this PR.

constanca-m:
I updated the workflow to only release a new tag if the new version in `version.py` is bigger than the latest released version. This way, if there was no change to the variable version in `version.py`, we also do not release a new tag, or if the version decreased somehow, we do not release it as well. @axw @gizas, is it better this way?

.github/workflows/release.yml
```diff
@@ -20,22 +20,55 @@ jobs:
...
+          # We will list all tags sorted by the version after the prefix lambda-v.
+          # We retrieve only the first line - that is, the most recent version.
+          # After that, we remove the prefix to only get the version number.
+          old_version=$(git tag --list --sort=-version:refname "lambda-v*" | awk 'NR==1{print $1}' | awk -F"lambda-v" '{ print $NF }')
+
+          # We now need to compare the current version inside version.py.
+          IFS='.' read -a new_numbers <<< ${{ env.VERSION }}
+          IFS='.' read -a old_numbers <<< $old_version
+
+          create_tag=false # only create tag if version increased
```
gizas, line 46:
```suggestion
          CREATE_TAG=false # only create tag if version increased
```

(I think you can use the final variable)

.github/workflows/release.yml
```diff
@@ -20,22 +20,55 @@ jobs:
...
+          for i in 0 1 2
+          do
+            if [[ ${new_numbers[i]} > ${old_numbers[i]} ]]
+            then
+              create_tag=true
```
gizas, line 51:
```suggestion
              CREATE_TAG==true
```

.github/workflows/release.yml
```diff
@@ -20,22 +20,55 @@ jobs:
...
+              break
+            elif [[ ${new_numbers[i]} < ${old_numbers[i]} ]]
+            then
+              break
+            fi
+          done
+
+          echo "CREATE_TAG=${create_tag}" >> $GITHUB_ENV
```
gizas, line 59:
```suggestion
          echo "CREATE_TAG=${CREATE_TAG}" >> $GITHUB_ENV
```

.github/workflows/release.yml
```diff
@@ -20,22 +20,55 @@ jobs:
...
+          old_version=$(git tag --list --sort=-version:refname "lambda-v*" | awk 'NR==1{print $1}' | awk -F"lambda-v" '{ print $NF }')
+
+          # We now need to compare the current version inside version.py.
+          IFS='.' read -a new_numbers <<< ${{ env.VERSION }}
+          IFS='.' read -a old_numbers <<< $old_version
+
+          create_tag=false # only create tag if version increased
+          for i in 0 1 2
+          do
+            if [[ ${new_numbers[i]} > ${old_numbers[i]} ]]
```
gizas, line 49: Does this compare only the minor version?

For example if version changes from 1.2.1 to 2.1.0 will it work?
(sorry for being lazy and not testing )

constanca-m: Yes, it iterates over every number between the `.`

gizas (approved):

axw (approved):
Looks good!

Issue #266, opened 2023-03-02
[Direct Deploy] Avoid using AWS Managed IAM policies with wildcard resources 


girodav:
(PLACEHOLDER: more details to be added soon)

For confirmed bugs, please report:
- Version:
- Steps to Reproduce:

singhvirohan:
@girodav since it's been a while since this issue was created, are you aware of any updates in regards to this ask? We've had a customer report this and are looking for clarifications on how to address this for them. 

https://github.com/elastic/security/issues/2007 has also been opened with the Security team.

PR #693, opened 2024-04-19
Use custom policy with restricted resource to replace managed policies

kaiyan-sheng:
## What does this PR do?

Elastic Serverless Forwarder Lambda Role uses the following AWS managed policies:

AWSLambdaSQSQueueExecutionRole
AWSLambdaKinesisExecutionRole
AWSLambdaBasicExecutionRole
AWSLambdaVPCAccessExecutionRole

All the above roles have attached policies with wildcard Resource statements (i.e "Resource": "*"). In order to limit the resource here, I'm replacing these policies with custom policies.

## Why is it important?

Security concern with IAM definition including too many wildcards.
https://github.com/elastic/elastic-serverless-forwarder/issues/266

## Checklist

### DIFF
--- a/publish_lambda.sh
+++ b/publish_lambda.sh
@@ -481,19 +481,108 @@ if __name__ == "__main__":
                 ],
             },
             "RoleName": "${CUSTOM_ROLE_PREFIX}ApplicationElasticServerlessForwarderRole",
-            "ManagedPolicyArns": [
-                "arn:${AWS_OR_AWS_GOV}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole",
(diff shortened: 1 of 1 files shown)


constanca-m:
I had a look again to try to understand where these policies come from, and I believe it depends on the way we are deploying ESF.

So if we use the [terraform files](https://github.com/elastic/terraform-elastic-esf) (private repo), we do not use managed policies like AWSLambdaSQSQueueExecutionRole, but instead grant the right permissions to the right resources. Exception is the use of policy AWSLambdaBasicExecutionRole. This would have to be changed as well.

For the publish script, you are already covering the cases in the changed file.

I don't think we are using managed policies anywhere else, but it is best to double check.

kaiyan-sheng:
I have not worked on this for a long time, will close it for now. If we decide to work on this issue again, I can reopen it.

Issue #721, opened 2024-05-16
Enable shared outputs and multiple outputs per input


constanca-m:
## Enhancement 

Currently, each input needs to have defined one output (see an example of the configuration file [here](https://www.elastic.co/guide/en/esf/current/aws-deploy-elastic-serverless-forwarder.html#sample-s3-config-file)).

However, this is not convenient. If a user wants to use the same output for 4 inputs, they have to define it 4 times. And if they want to change it later, they have to do it another 4 times, when we could place them as a shared output.

There are also two issues ([1](https://github.com/elastic/enhancements/issues/21474) and [2](https://github.com/elastic/elastic-serverless-forwarder/issues/704)) requesting the possibility to have more than one output defined for input. At this moment, this is not possible, unless the output type is different. Current options are:

https://github.com/elastic/elastic-serverless-forwarder/blob/acbe70242afad1d5061d64fd4d12b7e647de3768/share/config.py#L15

The outputs for each input are saved in a dictionary:

https://github.com/elastic/elastic-serverless-forwarder/blob/acbe70242afad1d5061d64fd4d12b7e647de3768/share/config.py#L294

However, this dictionary is using the type as the key, so if we try to add a new output with the same type, it will fail:

https://github.com/elastic/elastic-serverless-forwarder/blob/acbe70242afad1d5061d64fd4d12b7e647de3768/share/config.py#L421-L430

Additionally, and since in the code,each input is uniquely identified by its id (see [this function](https://github.com/elastic/elastic-serverless-forwarder/blob/acbe70242afad1d5061d64fd4d12b7e647de3768/share/config.py#L462)), I also do not think it is possible to configure the same input to have multiple outputs at all, even if specified in double.

To further clarify this change, let's consider this example: I want to:

1. Have two inputs, each for corresponding to one cloudwatch logs group. Both should send data to the same ES.
2. Have two outputs for one of the two cloudwatch logs group.

For this, the best we can do is:

```yaml
inputs:
  - id: "<ID-1>"
    type: cloudwatch-logs
    outputs:
      - args:
          api_key: "<API-KEY-1>"
          elasticsearch_url": "<URL-1>"
          es_datastream_name: "logs-esf.cloudwatch-default"
        type: elasticsearch
  - id: "<ID-2>"
    type: cloudwatch-logs
    outputs:
      - args:
          api_key: "<API-KEY-1>"
          elasticsearch_url": "<URL-1>"
[...]

PR #725, opened 2024-05-29
Enable multiple outputs for each input

constanca-m:
## What does this PR do?

Each input can now have multiple outputs of the same `type`. It cannot, however, have the same output specified more than once for each input - that is, it cannot have two `elasticsearch` outputs with the same destination. See section **Results** for examples.

## Why is it important?

See details on https://github.com/elastic/elastic-serverless-forwarder/issues/721.

## Checklist

Refer to https://github.com/elastic/elastic-serverless-forwarder/tree/main/how-to-test-locally.

## Related issues

Relates to https://github.com/elastic/elastic-serverless-forwarder/issues/721

### Example 1: Trying 2 `elasticsearch` outputs

I have an input with two `elasticsearch` outputs:
```yaml
"inputs":
- "id": "arn:aws:logs:eu-west-2:627286350134:log-group:constanca-test-local-esf:*"
  "outputs":
  - "args":
      "api_key": "..."
      "elasticsearch_url": "https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io"
      "es_datastream_name": "logs-esf.cloudwatch-default"
    "type": "elasticsearch"
  - "args":
      "api_key": "..."
      "elasticsearch_url": "https://terraform-2.es.europe-west4.gcp.elastic-cloud.com"
      "es_datastream_name": "logs-esf.cloudwatch-default"
    "type": "elasticsearch"
  "type": "cloudwatch-logs"
```

I sent a log event from this input.

If I look at Discover in both clouds, I can see that both outputs got my message:
- https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io:
Screenshot from 2024-05-29 14-19-22

- https://terraform-2.es.europe-west4.gcp.elastic-cloud.com:
Screenshot from 2024-05-29 14-19-51

### Example 2: Trying 2 `elasticsearch` outputs with the same destination

I have an input with two `elasticsearch` outputs that are the same:
```yaml
"inputs":
- "id": "arn:aws:logs:eu-west-2:627286350134:log-group:constanca-test-local-esf:*"
  "outputs":
  - "args":
      "api_key": "..."
      "elasticsearch_url": "https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io"
      "es_datastream_name": "logs-esf.cloudwatch-default"
    "type": "elasticsearch"
  - "args":
      "api_key": "..."
      "elasticsearch_url": "https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io"
      "es_datastream_name": "logs-esf.cloudwatch-default"
    "type": "elasticsearch"
  "type": "cloudwatch-logs"
```

This case fails, since we cannot have duplicated outputs (that is, output with the same destination for the same input):

```logs
[ERROR] ConfigFileException: An error occurred while applying output configuration at position 2 for input arn:aws:logs:eu-west-2:627286350134:log-group:constanca-test-local-esf:*: Duplicated output destination https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io for type elasticsearch
Traceback (most recent call last):
  File "/var/task/handlers/aws/utils.py", line 63, in wrapper
    return func(lambda_event, lambda_context)
  File "/var/task/handlers/aws/utils.py", line 98, in wrapper
    raise e
  File "/var/task/handlers/aws/utils.py", line 82, in wrapper
    return func(lambda_event, lambda_context)
  File "/var/task/handlers/aws/handler.py", line 74, in lambda_handler
    raise ConfigFileException(e)
```

### Example 3: Checking message body in replay queue

The message body caused by an ingestion error should contain the `output_destination` (instead of `output_type` like before). I am causing the placement of the message in the replay queue by using wrong authentication:

```yaml
"inputs":
- "id": "arn:aws:logs:eu-west-2:627286350134:log-group:constanca-test-local-esf:*"
  "outputs":
  - "args":
      "api_key": "<Wrong API Key"
      "elasticsearch_url": "https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io"
      "es_datastream_name": "logs-esf.cloudwatch-default"
    "type": "elasticsearch"
```

Message in the replay queue:

```json
{"output_destination":"https://terraform-8b3bac.es.eu-central-1.aws.cloud.es.io","output_args":{"es_datastream_name":"logs-esf.cloudwatch-default"},"event_payload":...
```

### DIFF
--- a/handlers/aws/replay_trigger.py
+++ b/handlers/aws/replay_trigger.py
@@ -20,7 +20,9 @@ class ReplayedEventReplayHandler:
     def add_event_with_receipt_handle(self, event_uniq_id: str, receipt_handle: str) -> None:
         self._events_with_receipt_handle[event_uniq_id] = receipt_handle
 
-    def replay_handler(self, output_destination: str, output_args: dict[str, Any], event_payload: dict[str, Any]) -> None:
+    def replay_handler(
+        self, output_destination: str, output_args: dict[str, Any], event_payload: dict[str, Any]
+    ) -> None:
         event_uniq_id: str = event_payload["_id"] + output_destination
         self._failed_event_ids.append(event_uniq_id)
 
--- a/handlers/aws/replay_trigger.py
+++ b/handlers/aws/replay_trigger.py
@@ -20,8 +20,8 @@ class ReplayedEventReplayHandler:
     def add_event_with_receipt_handle(self, event_uniq_id: str, receipt_handle: str) -> None:
         self._events_with_receipt_handle[event_uniq_id] = receipt_handle
 
-    def replay_handler(self, output_type: str, output_args: dict[str, Any], event_payload: dict[str, Any]) -> None:
-        event_uniq_id: str = event_payload["_id"] + output_type
+    def replay_handler(self, output_destination: str, output_args: dict[str, Any], event_payload: dict[str, Any]) -> None:
+        event_uniq_id: str = event_payload["_id"] + output_destination
         self._failed_event_ids.append(event_uniq_id)
 
     def flush(self) -> None:
--- a/handlers/aws/replay_trigger.py
+++ b/handlers/aws/replay_trigger.py
@@ -37,7 +37,7 @@ class ReplayedEventReplayHandler:
 
 def get_shipper_for_replay_event(
     config: Config,
-    output_type: str,
+    output_destination: str,
     output_args: dict[str, Any],
     event_input_id: str,
     replay_handler: ReplayedEventReplayHandler,
@@ -46,28 +46,28 @@ def get_shipper_for_replay_event(
     if event_input is None:
         raise InputConfigException(f"Cannot load input for input id {event_input_id}")
 
-    output: Optional[Output] = event_input.get_output_by_type(output_type)
+    output: Optional[Output] = event_input.get_output_by_destination(output_destination)
     if output is None:
-        raise OutputConfigException(f"Cannot load output of type {output_type}")
+        raise OutputConfigException(f"Cannot load output with destination {output_destination}")
 
     # Let's wrap the specific output shipper in the composite one, since the composite deepcopy the mutating events
     shipper: CompositeShipper = CompositeShipper()
 
-    if output_type == "elasticsearch":
+    if output.type == "elasticsearch":
         assert isinstance(output, ElasticsearchOutput)
         output.es_datastream_name = output_args["es_datastream_name"]
         shared_logger.debug("setting ElasticSearch shipper")
-        elasticsearch: ProtocolShipper = ShipperFactory.create_from_output(output_type=output_type, output=output)
+        elasticsearch: ProtocolShipper = ShipperFactory.create_from_output(output_type=output.type, output=output)
 
         shipper.add_shipper(elasticsearch)
         shipper.set_replay_handler(replay_handler=replay_handler.replay_handler)
 
         return shipper
 
-    if output_type == "logstash":
+    if output.type == "logstash":
         assert isinstance(output, LogstashOutput)
         shared_logger.debug("setting Logstash shipper")
-        logstash: ProtocolShipper = ShipperFactory.create_from_output(output_type=output_type, output=output)
+        logstash: ProtocolShipper = ShipperFactory.create_from_output(output_type=output.type, output=output)
 
         shipper.add_shipper(logstash)
         shipper.set_replay_handler(replay_handler=replay_handler.replay_handler)
--- a/share/config.py
+++ b/share/config.py
@@ -443,6 +443,7 @@ class Input:
             # for the same url/cloud_id for both types logstash or elasticsearch
             raise ValueError(f"Duplicated output destination {output_dest} for type {output_type}")
 
+        output: Optional[Output] = None
         if output_type == "elasticsearch":
             output = ElasticsearchOutput(**kwargs)
         elif output_type == "logstash":
--- a/share/config.py
+++ b/share/config.py
@@ -407,7 +407,6 @@ class Input:
 
         return list(self._outputs.keys())
 
-    # TODO
     def delete_output_by_destination(self, output_destination: str) -> None:
         """
         Output deleter.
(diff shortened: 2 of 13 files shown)


constanca-m:
I want to see if there is any significant change in performance between our latest released version, lambda-v1.14.0, and the ESF created from this PR.

To check this, I am going to trigger ESF based on messages coming to a cloudwatch logs group. This group is receiving some X number of events, then stops for 30s, and then the cycle repeats. The number of events doubles every 20 cycles. I let it run 100 cycles. Summarizing:

| Number of events sent + sleep 30s | Cycles |
|-----------------------------------|--------|
| 50                                | 20     |
| 100                               | 20     |
| 200                               | 20     |
| 400                               | 20     |
| 800                               | 20     |

The message being sent is small sized, only 94b.


I also want to check how long it takes to get a message from the cloudwatch groups to elasticsearch. Since ESF does not add to a document to ingest timestamp, I am going to add a default ingest pipeline and add that field in every document. 


<details>
<summary>You can add the ingest pipeline like this.</summary>

```
# Create pipeline to add ingest timestamp
PUT _ingest/pipeline/add_date
{
  "description" : "This pipeline add the ingest time to the document",
  "processors" : [
    {
      "set" : {
        "field": "received",
        "value": "{{_ingest.timestamp}}"
      }
    }
  ]
}

# Creates a component template for index settings
PUT _component_template/logs-esf-settings
{
  "template": {
    "settings": {
      "index.default_pipeline": "add_date",
      "index.lifecycle.name": "logs"
    }
  }
}

# Creates an index template
PUT _index_template/logs-esf-template
{
  "index_patterns": ["logs-esf.cloudwatch-*"],
  "data_stream": { },
  "priority": 500,
  "composed_of": ["logs-esf-settings"]
}

# Rollover the data stream to use an index with the pipeline
POST logs-esf.cloudwatch-default/_rollover
```
</details>


I deployed ESF and updated the lambda 3 times:
1. First run: ESF is using lambda-v1.14.0 and it has 1 `elasticsearch` output
2. Second run: ESF is using the code from this PR and it has 1 `elasticsearch` output
3. Third run: ESF is using the code from this PR and it has 2 `elasticsearch` outputs

So the first line in my graphics corresponds to first run, second line to second run, and third line to third run. The cloudwatch metrics for the ESF lambda function are these:

image

There is very little difference between the 3 ESFs and their configurations. For the 2 outputs configurations, the average duration slightly increased. This is to be expected, as now it needs to iterate over two outputs.

I also computed the average latency using the field my ingest pipeline added. I am calculating the latency this way:
1. I create a visualization graph with the formula `last_value(received)-last_value(@timestamp)`.
2. I use a minimum interval of `30s`. I am only using one document (the last one) for each interval.

The latency for 1 output (the one that is common for all three runs) is like this:
image

The latency for this output barely changed between my 3 runs.

The second output (third run) has a bigger latency:
image

This is to be expected as well. We can see that the minimum average latency is already past the latency of the first output. From this we can conclude that ESF sent the data to the first output first, and then it sent the data to this second output. They did not happen in parallel.

My conclusions is that there is no significant change or decay in performance. However, the more outputs the user adds, the bigger the latency gets for each output.

handlers/aws/replay_trigger.py
```diff
@@ -46,28 +48,28 @@ def get_shipper_for_replay_event(
...
+    output: Optional[Output] = event_input.get_output_by_destination(output_destination)
     if output is None:
-        raise OutputConfigException(f"Cannot load output of type {output_type}")
+        raise OutputConfigException(f"Cannot load output with destination {output_destination}")
 
     # Let's wrap the specific output shipper in the composite one, since the composite deepcopy the mutating events
     shipper: CompositeShipper = CompositeShipper()
 
-    if output_type == "elasticsearch":
+    if output.type == "elasticsearch":
```
gizas, line 58: I would suggest to either keep output.type or output_destination common in the whole doc

So output.type to become output.destination

share/config.py
```diff
@@ -394,40 +391,57 @@ def include_exclude_filter(self, value: IncludeExcludeFilter) -> None:
...
         """
         Output setter.
         Set an output given its type and init kwargs
         """
-        if not isinstance(output_type, str):
-            raise ValueError("`type` must be provided as string")
 
-        if output_type in self._outputs:
-            raise ValueError(f"Duplicated `type` {output_type}")
+        if output_type not in _available_output_types:
```
gizas, line 424: Same here to chaneg type with destination

kaiyan-sheng (approved):

handlers/aws/replay_trigger.py
constanca-m: `output.destination` is the URL - either the cloud url or ES url. `output.type` can only be `elasticsearch` or `logstash`, they are different things @gizas

share/config.py
constanca-m: Same comment as before, `output.type` can only be elasticsearch or logstash, we use it to create the shipper, not the same as the destination

share/config.py
```diff
@@ -394,40 +391,57 @@ def include_exclude_filter(self, value: IncludeExcludeFilter) -> None:
...
+                raise ValueError("Either `elasticsearch_url` or `cloud_id` must be set")
+            # elasticsearch_url takes precedence over cloud_id
+            if "elasticsearch_url" not in kwargs:
+                output_dest = kwargs["cloud_id"]
+            else:
+                output_dest = kwargs["elasticsearch_url"]
+        elif output_type == "logstash":
+            if "logstash_url" not in kwargs:
+                raise ValueError(f"Output type {output_type} requires logstash_url to be set")
+            output_dest = kwargs["logstash_url"]
```
constanca-m, line 439: @gizas You can see in this line how the output destination is initiated

handlers/aws/replay_trigger.py
gizas: Ok I had the impression that output.destination does not exist. Thanks for clarification
So destination should become output_url in general to denote the actual info of the variabble. But nw not important

gizas (approved):

PR #727, opened 2024-06-05
Removed '$' sign from region check in publish_lambda.sh

### LINKED ISSUE #726: Issue with region name check in publishing script
https://github.com/elastic/elastic-serverless-forwarder/blob/befdd23dca45624b10c18c7d17bc30a87e8d1d1a/publish_lambda.sh#L45

In the publishing script version referenced above, the `*"$gov"*` if statement evaluates to `**`, so it always returns S3 bucket ARN configured with "aws-us-gov" region.

I suggest remove `$` from the if statement -> `*"gov"*` or set up `gov="gov"` before it.

hvltaj:
Bug

## What does this PR do?

Fixed bug in the `publish_lambda.sh` script which always set up `gov` as bucket region.

## Why is it important?

Deployment script only works for `gov` regions in current state

## Checklist

- [~~] My code follows the style guidelines of this project
- [~~] I have commented my code, particularly in hard-to-understand areas
- [~~] I have made corresponding changes to the documentation
- [~~] I have made corresponding change to the default configuration files
- [~~] I have added tests that prove my fix is effective or that my feature works

Use deployment script to deploy Lambda.

## Related issues
Closes #726

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ### v1.17.1 - 2024/09/23
 ##### Bug fixes
 * Cache EC2:DescribeRegion API response to avoid throttling and improve performance [803](https://github.com/elastic/elastic-serverless-forwarder/pull/803).
-* Fixed deployment script bug that always set S3 region to `gov` [#727](https://github.com/elastic/elastic-serverless-forwarder/pull/727).
 
 ### v1.17.0 - 2024/07/10
 ##### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v1.15.0 - 2024/06/05
+##### Bug fixes
+* Fixed deployment script bug that always set S3 region to `gov` [#727](https://github.com/elastic/elastic-serverless-forwarder/pull/727).
+
 ### v1.14.0 - 2024/05/07
 ##### Bug fixes
 * Report misconfigured input ids as an error instead of warning, and place those messages in the replaying queue [#711](https://github.com/elastic/elastic-serverless-forwarder/pull/711).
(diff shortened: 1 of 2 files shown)


zmoog:
Hey @hvltaj, thanks for contributing the fix.

Adding this to the next release.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### v1.17.1 - 2024/09/17
```
constanca-m, line 1: This version should match the one inside `share/version.py`:

https://github.com/elastic/elastic-serverless-forwarder/blob/8be4fc4b87a06aa857e0cbc227f55f0533d89035/share/version.py#L5

Perhaps we should update the workflow to compare these, this should not pass...

constanca-m: But updating the `version.py` will also trigger a new release, which seems a bit unnecessary, since the dependencies are not changing (the dependencies can be seen [here](https://github.com/elastic/elastic-serverless-forwarder/blob/8be4fc4b87a06aa857e0cbc227f55f0533d89035/.github/workflows/upload-dependencies.yml#L55-L59)).

zmoog: Yeah, @constanca-m, you're right. 

@hvltaj, the `publish_lambda.sh` is not part of the release; it's a supporting tool, so we can avoid updating the `CHANGELOG.md` and `share/version.py` files.

kaiyan-sheng: I updated the changelog after https://github.com/elastic/elastic-serverless-forwarder/pull/803 is merged. Now it should match the version inside share/version.py.

kaiyan-sheng: ahh good point, ok I will take out the entry from CHANGELOG.md then!

kaiyan-sheng (approved):

constanca-m (approved):

CHANGELOG.md
constanca-m: I have created https://github.com/elastic/elastic-serverless-forwarder/issues/807 to see how we can stop this from happening, or maybe ease this version update process a bit.

PR #733, opened 2024-07-03
Add dead letter index for ES outputs

emilioalvap:
## What does this PR do?

Depends on https://github.com/elastic/elastic-serverless-forwarder/pull/729.
Closes https://github.com/elastic/enhancements/issues/21378.

Adds `es_dead_letter_index` option to ES outputs, with a similar mechanism as to mimic [filebeat's non_indexable_policy](https://www.elastic.co/guide/en/beats/filebeat/current/elasticsearch-output.html#_non_indexable_policy`).

After failure to ingest on a bulk() request, ES output will now try to ingest to the provided dead_letter_index before queueing the failed event on the replay queue.

## Checklist

### DIFF
--- a/shippers/es.py
+++ b/shippers/es.py
@@ -19,7 +19,7 @@ from .shipper import EventIdGeneratorCallable, ReplayHandlerCallable
 
 _EVENT_BUFFERED = "_EVENT_BUFFERED"
 _EVENT_SENT = "_EVENT_SENT"
-
+_VERSION_CONFLICT = 409
 
 class JSONSerializer(Serializer):
     mimetype = "application/json"
@@ -166,6 +166,11 @@ class ElasticsearchShipper:
             shared_logger.warning(
                 "elasticsearch shipper", extra={"error": error["create"]["error"], "_id": error["create"]["_id"]}
             )
+
+            if "status" in error["create"] and  error["create"]["status"] == 409:
+                    # Skip duplicate events on replay queue
+                    continue
+             
             shared_logger.debug("elasticsearch shipper", extra={"action": action_failed[0]})
             if self._replay_handler is not None:
                 self._replay_handler(self._output_destination, self._replay_args, action_failed[0])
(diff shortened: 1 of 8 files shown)


shippers/es.py
```diff
@@ -207,21 +214,87 @@ def send(self, event: dict[str, Any]) -> str:
...
+
+        # If no action can be encoded, return original action list as failed
+        if len(encoded_actions) == 0:
+            return dead_letter_errors
+
+        errors = es_bulk(self._es_client, encoded_actions, **self._bulk_kwargs)
+        failed = self._handle_outcome(actions=encoded_actions, errors=errors)
+
+        if not isinstance(failed, list) or len(failed) == 0:
+            return []
```
emilioalvap, line 265:
```suggestion
            return dead_letter_errors
```

gizas:
If you plan to merge it with 729 pr, please update changelog and include this PR - as per comment https://github.com/elastic/elastic-serverless-forwarder/pull/729#issuecomment-2217632367

constanca-m:
@emilioalvap I saw you tested the situation for when both data streams fail with a processor. Does it also happen if the dead letter index does not exist or if we pass an empty string ` ` as a value for it?

emilioalvap:
@constanca-m As for an index that doesn't exist, that is usually handled by ES [auto create policy](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-creation), which tbh it's not something I've ever come across to be disabled. 
Even then with auto_create disabled, my understanding is that the `_bulk` request will return an index not found response (404) and report it back as an error for each individual action, like a mapping error or processor rejection would. So it would behave the same way as those types of errors and end up in the replay queue.

As for the empty value, it's the default value if not defined, so it should skip the dead letter index and continue straight into replay queue, as it is now.

constanca-m (approved):

gizas (approved):

Issue #779, opened 2024-08-28
dead letter index: store errors using the ECS field format


zmoog:
## Situation

Starting from version v1.17.0, ESF sends documents with indexing errors to the dead letter index if users set an index/data stream name in the `es_dead_letter_index` config option.

ESF encodes the document using this format [^1]:

```python
        return {
            "@timestamp": datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
            "_id": f"{uuid.uuid4()}",
            "_index": self._es_dead_letter_index,
            "_op_type": "create",
            "message": json_dumper(outcome["action"]),
            "error": outcome["error"],
        }
```

The `error` field value is a `str` from the `elastic search.helpers.bulk` response.

## Problem

A user reported mapping conflicts on their dead letter index's `error` field. They expected the `error` field to be an object and to adopt the ECS format at https://www.elastic.co/guide/en/ecs/current/ecs-error.html

## Solutions

Switch to the ECS error format to encode documents with index failure.

[^1]: https://github.com/elastic/elastic-serverless-forwarder/blob/main/shippers/es.py#L290

zmoog:
Here is an example of an actual error value:

![Image](https://github.com/user-attachments/assets/e43fcf70-594d-4522-8593-39b463a7613a)

The second return value is a list of dicts with the following keys:

| Header | Header |Header |
|--------|--------|--------|
| `error` | `str` | ConnectionError(<urllib3.connection.HTTPConnection object at 0x1064237f0>: Failed to establish a new connection: [Errno 61] Connection refused) caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x1064237f0>: Failed to establish a new connection: [Errno 61] Connection refused) |
| `status` | `str` | 'N/A' |
| `exception` | `ConnectionError` | |
| `data` | `dics` | {'field1': 'test value 0', 'field2': 'another test value 0'} |
| `_index` | `str` | 'test-index' |

PR #793, opened 2024-09-12
dead letter index: align error field to ECS and do not forward retryable errors

kaiyan-sheng:
## What does this PR do?

This PR brings the following changes:

- Align the `error` field in documents sent to the dead letter index (DLI) to the ECS format; the field now provides `error.message` and `error.type`.
- Add the `http.response.status_code` field 
- Limit the error type sent to the DLI; do not send Elasticsearch client errors that:
  - are connection errors (do not have an `http.response.status_code`)
  - have a retriable status code (for example, 429)

Here is a sample error document from a mapping conflict:

```json
{
  "@timestamp": "2024-10-07T05:57:59.448925Z",
  "message": "{\"hey\":{\"message\":\"hey there\"},\"_id\":\"e6542822-4583-438d-9b4d-1a3023b5eeb9\",\"_op_type\":\"create\",\"_index\":\"logs-succeed.pr793-default\"}",
  "error": {
    "message": "[1:30] failed to parse field [hey] of type [keyword] in document with id 'e6542822-4583-438d-9b4d-1a3023b5eeb9'. Preview of field's value: '{message=hey there}'",
    "type": "document_parsing_exception"
  },
  "http": {
    "response": {
      "status_code": 400
    }
  }
}

```

## Why is it important?

- The error field must always match the definition of the ECS error field.
- Avoid sending connection-related errors to DLI.
- Avoid sending errors with retryable status codes to DLI.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-serverless-forwarder/issues/779


zmoog:
@constanca-m, I created https://github.com/elastic/elastic-serverless-forwarder/pull/799 as an experiment, I guess we can focus on this one.

zmoog:
I'm running the `ElasticsearchShipper` in different scenarios to see what errors we get from the Elasticsearch client.

## Invalid elasticsearch_url 

Errors when the `elasticsearch_url` points to a server:port where no server is running:

CleanShot 2024-09-17 at 17 14 06@2x

## Mapping exception

The document failed due to a mapping exception:

CleanShot 2024-09-17 at 17 12 08@2x

zmoog:
We have different fields depending on the error type, and some fields can have various types (`status` can be an `str` or an `int`, and `error` can be an `str` or a `dict`).

zmoog:
Quick recap:

| Field | 5xx | N/A |
|--------|--------|--------|
| `status` | `int` | `str` |
| `error` | `dict` |  `str`|
| `exception` |  |  `class` |

constanca-m:
> We have different fields depending on the error type, and some fields can have various types (status can be an str or an int, and error can be an str or a dict).

Nice research @zmoog ! In that case we should include a test in the code with the possible different types.

zmoog:
We are targeting the following mappings:

```json
{
  "mappings": {
    "dynamic": "false",
    "_data_stream_timestamp": {
      "enabled": true
    },
    "date_detection": false,
    "numeric_detection": false,
    "properties": {
      "error": {
        "properties": {
          "code": {
            "type": "keyword",
            "ignore_above": 1024
          },
          "id": {
            "type": "keyword",
            "ignore_above": 1024
          },
          "message": {
            "type": "match_only_text"
          },
          "stack_trace": {
            "type": "wildcard",
            "fields": {
              "text": {
                "type": "match_only_text"
              }
            }
          },
          "type": {
            "type": "keyword",
            "ignore_above": 1024
          }
        }
      }
    }
  }
}

```

zmoog:
We should probably at least provide a `message`:

Here's a possible mapping between error conditions and the `error` field to post to the dead letter index:

| Field | 5xx | N/A |
|--------|--------|--------|
| `message ` | From `error.reason`? | From `error`? |
| `type ` | From `error.type`? | From `type(exteption)` or similar? |

When available, we may also consider adding a [http.response.status_code](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-response-status-code) with the status code.

Dumping all the options now, don't consider this a plan (yet).

kaiyan-sheng:
@zmoog Thank you for the comparisons!! What's the `exception.info` struct in N/A case? Seems like that one has some information we can use for the N/A case.

zmoog:
> What's the `exception.info` struct in N/A case? Seems like that one has some information we can use for the N/A case.

The `errors[0]['create']['exception']` is a `elasticsearch.exceptions.ConnectionError` instance:

CleanShot 2024-09-26 at 17 45 28@2x

shippers/es.py
```diff
@@ -176,7 +176,21 @@ def _handle_outcome(self, actions: list[dict[str, Any]], errors: tuple[int, Unio
                 # Skip duplicate events on dead letter index and replay queue
                 continue
 
-            failed.append({"error": error["create"]["error"], "action": action_failed[0]})
+            failed_error = {"action": action_failed[0]} | self._parse_error(error["create"])
+
+            # error_field = error.get("create", {}).get("error", None)
```
kaiyan-sheng, line 181: should we remove all these?

shippers/es.py
```diff
@@ -185,6 +199,52 @@ def _handle_outcome(self, actions: list[dict[str, Any]], errors: tuple[int, Unio
...
+            # error field, if present, and the type is an integer.
+            #
+            # Sometimes the status code is a string, for example,
+            # when the connection to the server fails.
+            field["http"] = {"response": {"status_code": error["status"]}}
+
+        if "error" not in error:
+            return field
+
+        if isinstance(error["error"], str):
```
kaiyan-sheng, line 231: should we add tests for both str and dict cases?

shippers/es.py
zmoog: Removed!

zmoog:
> @zmoog Thanks! I cant approve it but it looks good to me.

Adding a few more unit tests and fixing the integration tests.

shippers/es.py
zmoog: I added unit tests for the  `_parse_error()` function.

emilioalvap (dismissed):
LGTM

constanca-m (dismissed):

zmoog:
For the records, I tried sending the following test document:

```json
{
  "hey": {
      "message": "hey there",
  }
}
```

On a data stream where the field mapping is:

```json
{
  "hey": {
    "type": "keyword",
    "ignore_above": 1024
  }
}
```

Which causes a `document_parsing_exception`:

CleanShot 2024-10-02 at 10 25 54@2x

zmoog:
Hey reviewers! Sorry for the noise; I just pushed the last commit with the docs update 🙇

docs/en/aws-deploy-elastic-serverless-forwarder.asciidoc
```diff
@@ -230,6 +245,10 @@ For `elasticsearch` the following arguments are supported:
...
   * `args.es_dead_letter_index`: Name of data stream or index where logs should be redirected to, in case indexing to `args.es_datastream_name` returned an error.
+  * `es_dead_letter_forward_errors`: List of errors that should be forwarded to the dead letter index. The default value is an empty list (forward all errors). If the list is not empty, only the errors in the list will be forwarded to the dead letter index.
+    In general, you can use values from the `error.type` field in the Elasticseach API response. Here is a few examples of errors that can be forwarded to the dead letter index:
```
constanca-m, line 249: > In general, you can use values from the `error.type`

What does this mean exactly? I see the condition to not be sent is:

```python
(  # no http status: connection error
                self._es_dead_letter_forward_errors
                and action["error"]["type"] not in self._es_dead_letter_forward_errors
            ):
```

I don't really understand the first part of the condition. It means you have to write the whole error to exclude it? Is there another way to exclude an error other than its `error.type`? @zmoog

zmoog: Sorry, I had the feeling this part wasn't clear enough.

> In general, you can use values from the `error.type`

The `error.type` in this sentence refers to the Elasticsearch API response.

For example, here's the Elasticsearch API response for 

```json
{
  "error": {
    "root_cause": [
      {
        "type": "fail_processor_exception",
        "reason": "Fail message"
      }
    ],
    "type": "fail_processor_exception",
    "reason": "Fail message"
  },
  "status": 500
}
```

`_parse_error()` picks `error.type` from the ES API response, so the shipper can later compare it with the values in the `_es_dead_letter_forward_errors` list.

> I don't really understand the first part of the condition. It means you have to write the whole error to exclude it? 

I don't get what you mean with "you have to write the whole error to exclude it". 

> Is there another way to exclude an error other than its error.type?

Users reported problems with the DLI forwarding all errors to ES. 

This PR focuses on adding essential filtering capability:

- do not forward connection-related errors (we assumed errors without `http.response.status_code` are related o connection failures).
- filter out errors based on their `error.type`

There are probably more ways to filter the errors based on other criteria. 

However, the Failure Store (FS) will ship in one upcoming ES release. So, it's better to give ESF what it needs today without re-implementing Failure Store features.

zmoog: @constanca-m I tried to clarify the `es_dead_letter_forward_errors` option docs: does it flow better?

constanca-m (approved):

docs/en/aws-deploy-elastic-serverless-forwarder.asciidoc
zmoog: I will never ignore the feeling that a sentence isn't working again.

Thanks for raising the issue.

docs/en/aws-deploy-elastic-serverless-forwarder.asciidoc
```diff
@@ -230,6 +245,10 @@ For `elasticsearch` the following arguments are supported:
...
   * `args.es_dead_letter_index`: Name of data stream or index where logs should be redirected to, in case indexing to `args.es_datastream_name` returned an error.
+  * `es_dead_letter_forward_errors`: List of errors that should be forwarded to the dead letter index. The default value is an empty list (forward all errors). If the list is not empty, only the errors in the list will be forwarded to the dead letter index.
+    You can use the error types from the Elasticseach API (see `error.type` in API response). Here is a few examples:
```
kaiyan-sheng, line 249: Is there a link we can give the users for error types from ES API?

zmoog: I searched and failed. Let me try one more time before wrapping this up!

docs/en/aws-deploy-elastic-serverless-forwarder.asciidoc
emilioalvap: @zmoog Can we add the clarification that `error.type` is the one returned by `bulk()` api? AFAIK, there's no exhaustive list of error types but that should be a good starting point

zmoog:
@kaiyan-sheng, @constanca-m, @emilioalvap, I'm sorry for pushing yet another commit to this PR 🙇 

Our tech leads asked to simplify the criteria to forward to DLI along these lines. 

ESF should not forward the error to DLI when:

- there is no response from ES (not status code, probably due to connection errors)
- the status code from ES is retriable; for example, 429

constanca-m:
> there is no response from ES (not status code, probably due to connection errors)
> the status code from ES is retriable; for example, 429

Sounds good!

shippers/es.py
```diff
@@ -243,26 +295,56 @@ def flush(self) -> None:
...
+
+        Returns:
+            list[Any]: A list of actions that were not indexed in the DLI.
+        """
+        non_indexed_actions: list[Any] = []
         encoded_actions = []
-        dead_letter_errors: list[Any] = []
+
         for action in actions:
+            if (
```
constanca-m, line 319: :+1:

shippers/es.py
```diff
@@ -243,26 +295,56 @@ def flush(self) -> None:
...
+        1. The action response does not have an HTTP status (e.g., the connection failed).
+        2. The list of action errors to forward is not empty, and the action error type is not in the list.
+        3. The action could not be encoded for indexing in the DLI.
+        4. The action failed indexing attempt in the DLI.
+
+        Args:
+            actions (list[Any]): A list of actions to be processed.
+
+        Returns:
+            list[Any]: A list of actions that were not indexed in the DLI.
```
constanca-m, line 313: Why do we want to return the list of actions that were not indexed? I can't find its use in the code, but I might be missing it.

constanca-m:
@zmoog The documentation is no longer in this PR, did you remove it on purpose?

shippers/es.py
zmoog: The `_send_dead_letter_index()` aims to index errors in the DLI.

However, it may not be able to index errors for different reasons:

1. the function fails to encode the error
2. ES returns an error on the DLI indexing request
3. the error is retryable

So, the function returns the list of errors to the caller, saying, "These are the errors I could not (or won't) index on the DLI."

In this case, the caller (`flush()` function ) will send these errors to the replay queue.

The function structure didn't change. I added reason 3 ("the error is retryable") and opted for a more generic `non_indexed_actions` variable name instead of `dead_letter_errors` due to the change in semantics (we don't index errors only due to problems but also as a deliberate decision based on retriability).

zmoog:
> The documentation is no longer in this PR, did you remove it on purpose?

I removed it on purpose because the error types list no longer exists. But I should update the docs for the `es_dead_letter_index`, on it!

shippers/es.py
constanca-m: > and opted for a more generic non_indexed_actions variable name instead of dead_letter_errors

I see :sweat_smile: I thought now these errors were meant to be ignored, hence why I got confused

constanca-m (approved):

docs/en/aws-deploy-elastic-serverless-forwarder.asciidoc
zmoog: UPDATE: the team decided not to add additional config option, and change the default DLI behavior to not send (Elasticsearch client) errors that:

- are connection errors (do not have an `http.response.status_code`)
- have a retriable status code (for example, 429)

Issue #723, opened 2024-05-22
Investigate number of API calls of ESF due to ec2:describe_regions usage


gizas:
**Describe the enhancement:**
Currently in order to extract the region of cloudwatch logs_group stream ,we make use of ec2 DescribeRegions function in [here](https://github.com/elastic/elastic-serverless-forwarder/blob/main/handlers/aws/utils.py#L397). There have been cases reported where the number of concurrent lamda functions can trigger multiple times the DescribeRegions fucntion and raise the number of API calls until throttling events to happen.

This issue will try to investigate alternative solutions in order to overcome the API throttling in AWS

**Describe a specific use case for the enhancement or feature:**

- User created 50 lamda fucntions and this raised following issue: `"message": "An error occurred (RequestLimitExceeded) when calling the DescribeRegions operation (reached max retries: 4): Request limit exceeded."`

Possible solutions can be:
- User to provide the region list that we need to get information and make the describe_regions parametrised with specific list
- [describe_log_streams](https://docs.aws.amazon.com/cli/latest/reference/logs/describe-log-streams.html) on the config list that user has provided. This has limit of 5 concurrent requests

constanca-m:
When the function with this part of the code:

https://github.com/elastic/elastic-serverless-forwarder/blob/acbe70242afad1d5061d64fd4d12b7e647de3768/handlers/aws/utils.py#L397

is called, it means ESF was triggered by a cloudwatch logs group. I don't think a cloudwatch logs group can be in a different region than ESF. Unfortunately, I cannot find documentation on this, but I have tried using a different region and did not succeed. So I am not sure. But if this is correct, than getting the region should be very simple as the only thing we have to do is this:

```python
import os
  region = os.environ['AWS_REGION']
```

Reference [here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html).

constanca-m:
I think what I suggested in https://github.com/elastic/elastic-serverless-forwarder/issues/723#issuecomment-2124827063 would work for all types of deployment. But as I mentioned, I don't know if lambda can communicate with any resources in a different region - I think cloudwatch logs needs to be in the same one. If that is the case, that could be a solution. @bturquet

constanca-m:
I wanted to see the details of what is happening exactly in our code and why this is an issue. I wrote this while I was testing it:

The problem comes from this line:

[...]

PR #803, opened 2024-09-21
Cache describe_regions using lru_cache from stdlib

zmoog:
## What does this PR do?

Caches EC2:DescribeRegion API calls response.

## Why is it important?

On high-volume deployments, ESF can hit the EC2:DescribeRegions API requests limit, causing throttling errors like the following:

```text
An error occurred (RequestLimitExceeded) when calling the DescribeRegions operation (reached max retries: 4): Request limit exceeded.
```

ESF needs the list of existing regions to parse incoming events from the `cloudwatch-logs` input. Since new AWS region additions do not happen frequently, picking up and caching the list of existing regions at function startup seems adequate.

The list of existing AWS regions is available at https://aws.amazon.com/about-aws/global-infrastructure/regions_az/

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v1.17.1 - 2024/09/23
+##### Bug fixes
+* Cache EC2:DescribeRegion API response to avoid throttling and improve performance [803](https://github.com/elastic/elastic-serverless-forwarder/pull/803).
+
 ### v1.17.0 - 2024/07/10
 ##### Features
 * Add dead letter index for ES outputs [733](https://github.com/elastic/elastic-serverless-forwarder/pull/733).
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -387,7 +387,7 @@ def get_account_id_from_arn(lambda_arn: str) -> str:
 
 
 @lru_cache(maxsize=1)
-def describe_regions(all_regions: bool = True) -> dict[str, Any]:
+def describe_regions(all_regions: bool = True) -> Any:
     """
     Fetches all regions from AWS and returns the response.
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -2,8 +2,8 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 import os
-from typing import Any, Callable, Optional
 from functools import lru_cache
+from typing import Any, Callable, Optional
 
 import boto3
 from aws_lambda_typing import context as context_
@@ -390,7 +390,7 @@ def get_account_id_from_arn(lambda_arn: str) -> str:
 def describe_regions(all_regions: bool = True) -> dict[str, Any]:
     """
     Fetches all regions from AWS and returns the response.
-    
+
     :return: The response from the describe_regions method
     """
     return get_ec2_client().describe_regions(AllRegions=all_regions)
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -386,7 +386,7 @@ def get_account_id_from_arn(lambda_arn: str) -> str:
     return arn_components[4]
 
 
-@lru_cache(maxsize=1)
+@lru_cache()
 def describe_regions(all_regions: bool = True) -> Any:
     """
     Fetches all regions from AWS and returns the response.
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -406,13 +406,9 @@ def get_input_from_log_group_subscription_data(
     In order to not hardcode the list of regions we rely on ec2 DescribeRegions - as much weird as it is - that I found
     no information about having any kind of throttling. We add IAM permissions for it in deployment.
     """
-    # all_regions = get_ec2_client().describe_regions(AllRegions=True)
     all_regions = describe_regions(all_regions=True)
     assert "Regions" in all_regions
     for region_data in all_regions["Regions"]:
-
-        # arn:aws:logs:region:account-id:log-group:log_group_name:*
-
         region = region_data["RegionName"]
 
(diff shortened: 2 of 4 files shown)


zmoog:
I can probably add cache expiration to avoid a stale region list:

```python
import time
from functools import wraps
from typing import Any, Callable

def cache_for(seconds: int) -> Callable:
    def decorator(func: Callable) -> Callable:
        cache = {}
        stats = {'hits': 0, 'misses': 0}
        
        @wraps(func)
        def wrapper(*args, **kwargs) -> Any:
            current_time = time.time()
            if 'result' in cache and (current_time - cache['time'] < seconds):
                stats['hits'] += 1
                return cache['result']
            result = func(*args, **kwargs)
            cache['result'] = result
            cache['time'] = current_time
            stats['misses'] += 1
            return result
        
        wrapper.cache_stats = stats
        return wrapper
    return decorator

@cache_for(seconds=60)
def describe_regions(all_regions: bool = True) -> Any:
    """
    Fetches all regions from AWS and returns the response.

    :return: The response from the describe_regions method
    """
    return get_ec2_client().describe_regions(AllRegions=all_regions)

# Example usage with AWS regions
import boto3

def get_ec2_client():
    return boto3.client('ec2')

# Example usage to access cache statistics
if __name__ == "__main__":
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 0, 'misses': 1}

    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 1, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 2, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 3, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 4, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 5, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 6, 'misses': 1}
    describe_regions()
    print(describe_regions.cache_stats)  # {'hits': 7, 'misses': 1}
```

I get the following output:

```
{'hits': 0, 'misses': 1}
{'hits': 1, 'misses': 1}
{'hits': 2, 'misses': 1}
{'hits': 3, 'misses': 1}
{'hits': 4, 'misses': 1}
{'hits': 5, 'misses': 1}
{'hits': 6, 'misses': 1}
{'hits': 7, 'misses': 1}
```

zmoog:
Or, we can use a 3rd party library like https://cachetools.readthedocs.io/en/latest/

```python
from typing import Any

from cachetools.func import ttl_cache


@ttl_cache(ttl=1800) # 30 minutes
def describe_regions(all_regions: bool = True) -> Any:
    """
    Fetches all regions from AWS and returns the response.

    :return: The response from the describe_regions method
    """
    print("Fetching regions from AWS...")
    return get_ec2_client().describe_regions(AllRegions=all_regions)

# Example usage with AWS regions
import boto3

def get_ec2_client():
    return boto3.client('ec2')

# Example usage to access cache statistics
if __name__ == "__main__":
    
    # print(dir(describe_regions))

    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 0, 'misses': 1}

    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 1, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 2, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 3, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 4, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 5, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 6, 'misses': 1}
    describe_regions(all_regions=False)
    print(describe_regions.cache_info())  # {'hits': 7, 'misses': 1}
```

zmoog:
Thanks for the in-depth analysis.

I tested the cloudwatch lambda trigger on the AWS console and ESF. As of today, it seems cloudwatch lambda triggers can only work with log groups in the same region from as the lambda functions. For example, if I deploy ESF on `eu-west-1`, I can only process log events on log groups on `eu-west-1`.

Given this limit, there is no reason to keep calling the EC2:DescribeRegion API on every event.

I plan to remove this API call from ESF.

Here's my plan:

- Add the `@lru_cache` decorator from the standard library to reduce the number of API calls from one every event to just one on start. This is a small risk change that would allow to ship a patch release today.
- Go through the process of removing the API call (change the code to use the region from the function, remove the required permissions from the infrastructure, and test the whole package).

WDYT?

constanca-m:
I am fine with approving the PR as it is. 

You need to change the version of ESF currently (I believe you need to update the changelog and [version.py](https://github.com/elastic/elastic-serverless-forwarder/blob/main/share/version.py). After that the release workflow will be triggered, but if we push this change just like this, then nothing will happen.

zmoog:
> You need to change the version of ESF currently (I believe you need to update the changelog and [version.py](https://github.com/elastic/elastic-serverless-forwarder/blob/main/share/version.py). After that the release workflow will be triggered, but if we push this change just like this, then nothing will happen.

Thanks! On it.

constanca-m (approved):
Thanks @zmoog ! I would say we should keep https://github.com/elastic/elastic-serverless-forwarder/issues/723 open. I have copied and pasted my comment above so we know where we stand now.

I doubt we will have issues like this again with the cache, but let's see!

zmoog:
> I would say we should keep #723 open. 

I agree!

> I doubt we will have issues like this again with the cache, but let's see!

I'll work on removing the EC2:DescribeRegions API call later this week.

PR #811, opened 2024-09-24
Remove describe regions

zmoog:
## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/handlers/aws/handler.py
+++ b/handlers/aws/handler.py
@@ -34,6 +34,7 @@ from .utils import (
     get_shipper_from_input,
     get_sqs_client,
     get_trigger_type_and_config_source,
+    parse_arn,
     wrap_try_except,
 )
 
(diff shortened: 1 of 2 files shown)


handlers/aws/handler.py
```diff
@@ -141,6 +141,7 @@ def lambda_handler(lambda_event: dict[str, Any], lambda_context: context_.Contex
 
     if trigger_type == "cloudwatch-logs":
         cloudwatch_logs_event = _from_awslogs_data_to_event(lambda_event["awslogs"]["data"])
+        lambda_arn = parse_arn(lambda_context.invoked_function_arn)
```
constanca-m, line 144: Why do `parse_arn`?

We could just remove this line and the new function and do:

```python
region = os.environ['AWS_REGION']
```

[Here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime) are the lambda enviroment variables.

zmoog: I am not a fan of looking up environment variables in the middle of business code. I usually prefer to look them up during application initialization.

I see many uses of the ARN namedtouple, but I you're right: let's keep this simple.

zmoog: Yeah, I see. 

I', adding a helper function to deault with AWS_REGION and AWS_DEFAULT_REGION.

zmoog: Switched lambda region look up to environment variables.

constanca-m (approved):

PR #860, opened 2024-11-29
[s3-sqs] Log a Lambda event summary on exception

zmoog:
## What does this PR do?

If there is an exception when processing a lambda event from the `s3-sqs` input, ESF summarizes the lambda event and adds the summary as extra information to the exception log message.

The summary contains the essential elements for the lambda event. For the `s3-sqs` input, it includes the bucket ARN and the object key. The summary limits the number of records to `max_records`.

ESF currently supports lambda event summary for `s3-sqs` only. We want to hear users' feedback before making additional changes or extending the summary to other input types.

## Why is it important?

For some error classes (e.g., encoding errors), knowing the S3 objects in flight in the lambda event helps speed up investigation and troubleshooting.

Users can achieve the same goal by enabling APM, but this extra log helps those users who do not or won't use APM for any reason.

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files~~
- [ ] I have added an entry in `CHANGELOG.md` and updated `share/version.py`, if my change requires a new release.

Deploy ESF and send an SQS message with a S3 that points to a non-existent S3 object:

```sh
aws sqs send-message \
    --queue-url https://sqs.eu-west-1.amazonaws.com/627286350134/mbranca-esf-notifications \
    --message-body '{"Records":[{"awsRegion":"eu-west-1","eventTime": "2024-01-31T19:06:30.680Z","eventName":"ObjectCreated:Put","eventSource":"aws:s3","eventVersion":"2.1","s3":{"bucket":{"arn":"arn:aws:s3:::mbranca-esf-data","name":"mbranca-esf-data"},"object":{"key":"AWSLogs/123456/I-do-not-exist"}}}]}'
```

## Logs

Here is a sample log with the lambda event summary:

```json
{
    "@timestamp": "2024-11-29T12:06:05.404Z",
    "log.level": "error",
    "message": "exception raised",
    "ecs": {
        "version": "1.6.0"
    },
    "error": {
        "message": "An error occurred (404) when calling the HeadObject operation: Not Found",
        "stack_trace": "  File \"/var/task/handlers/aws/utils.py\", line 103, in wrapper\n    return func(lambda_event, lambda_context)\n  File \"/var/task/handlers/aws/handler.py\", line 532, in lambda_handler\n    for es_event, last_ending_offset, last_event_expanded_offset, current_s3_record in _handle_s3_sqs_event(\n  File \"/var/task/handlers/aws/s3_sqs_trigger.py\", line 141, in _handle_s3_sqs_event\n    for log_event, starting_offset, ending_offset, event_expanded_offset in events:\n  File \"/var/task/storage/s3.py\", line 78, in get_by_lines\n    s3_object_head = self._s3_client.head_object(Bucket=self._bucket_name, Key=self._object_key)\n  File \"/var/task/botocore/client.py\", line 565, in _api_call\n    return self._make_api_call(operation_name, kwargs)\n  File \"/var/task/elasticapm/instrumentation/packages/base.py\", line 213, in call_if_sampling\n    return self.call(module, method, wrapped, instance, args, kwargs)\n  File \"/var/task/elasticapm/instrumentation/packages/botocore.py\", line 106, in call\n    result = wrapped(*args, **kwargs)\n  File \"/var/task/botocore/client.py\", line 1017, in _make_api_call\n    raise error_class(parsed_response, operation_name)\n",
        "type": "ClientError"
    },
    "log": {
        "logger": "root",
        "origin": {
            "file": {
                "line": 128,
                "name": "utils.py"
            },
            "function": "wrapper"
        },
        "original": "exception raised"
    },
    "process": {
        "name": "MainProcess",
        "pid": 8,
        "thread": {
            "id": 139632879478592,
            "name": "MainThread"
        }
    },
    "service": {
        "name": "mbranca-esf-ApplicationElasticServerlessForwarder-0VuQ7x2yawGZ"
    },
    "span": {
        "id": "4a348c93af0ca023"
    },
    "summary": {
        "aws:sqs": {
            "records": [
                {
                    "bucket": {
                        "arn": "arn:aws:s3:::mbranca-esf-data",
                        "name": "mbranca-esf-data"
                    },
                    "object": {
                        "key": "AWSLogs/123456/I-do-not-exist"
                    }
                }
            ],
            "total_records": 1
        }
    },
    "trace": {
        "id": "88556e9d3f65fc709e53555b64b44a66"
    },
    "transaction": {
        "id": "c8dc45b175c39628"
    }
}
```

### DIFF
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -142,7 +142,7 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
     """
     Summarize the lambda event to include only the most relevant information.
     """
-    summary = {}
+    summary: dict[str, Any] = {}
 
     try:
         records = event.get("Records", [])
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -203,11 +203,11 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
                 # Update the summary with the new information.
                 summary["aws:sqs"] = aws_sqs_summary
 
-    except Exception as e:
-        shared_logger.exception("error summarizing lambda event", exc_info=e)
+    except Exception as exc:
+        shared_logger.exception("error summarizing lambda event", exc_info=exc)
         # We add an error message to the summary so users know if the summary
         # is incomplete.
-        summary["error"] = str(e)
+        summary["error"] = str(exc)
 
     return summary
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -145,6 +145,7 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
     summary: dict[str, Any] = {}
 
     try:
+        first_records_key = f"first_{max_records}_records"
         records = event.get("Records", [])
 
         for record in records:
@@ -153,22 +154,23 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
             if event_source == "aws:sqs":
                 aws_sqs_summary = summary.get(
                     "aws:sqs",
-                    # if the key does not exist, we initialize the summary
+                    # if `aws:sqs` key does not exist yet, 
+                    # we initialize the summary.
                     {
                         "total_records": 0,
-                        "records": [],
+                        first_records_key: [],
                     },
                 )
 
-                # The body contains the SQS message payload which is the S3
-                # notification event encoded as a JSON string.
-                event = json_parser(record["body"])
+                # We keep track of the total number of notifications in the
+                # lambda event, so users know if the summary is incomplete.
+                notifications = json_parser(record["body"])
 
                 # So users know if we included only a
                 # subset of the records.
-                aws_sqs_summary["total_records"] += len(event["Records"])
+                aws_sqs_summary["total_records"] += len(notifications["Records"])
 
-                for r in event["Records"]:
+                for r in notifications["Records"]:
                     # we only include the s3 object key in the summary.
                     #
                     # Here is an example of a record:
@@ -193,15 +195,19 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
                     #   ]
                     # }
 
-                    # We limit the number of records to `max_records` to
-                    # avoid large log payloads.
-                    if len(aws_sqs_summary["records"]) < max_records:
-                        aws_sqs_summary["records"].append(r.get("s3"))
+                    # We stop adding records to the summary once we reach
+                    # the `max_records` limit.
+                    if len(aws_sqs_summary[first_records_key]) < max_records:
+                        aws_sqs_summary[first_records_key].append(r.get("s3"))
 
+                # Update the summary with the new information.
                 summary["aws:sqs"] = aws_sqs_summary
 
     except Exception as e:
         shared_logger.exception("error summarizing lambda event", exc_info=e)
+        # We add an error message to the summary so users know if the summary
+        # is incomplete.
+        summary["error"] = str(e)
 
     return summary
 
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -154,7 +154,7 @@ def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict
             if event_source == "aws:sqs":
                 aws_sqs_summary = summary.get(
                     "aws:sqs",
-                    # if `aws:sqs` key does not exist yet, 
+                    # if `aws:sqs` key does not exist yet,
                     # we initialize the summary.
                     {
                         "total_records": 0,
--- a/handlers/aws/utils.py
+++ b/handlers/aws/utils.py
@@ -125,11 +125,85 @@ def wrap_try_except(
             if apm_client:
                 apm_client.capture_exception()
 
-            shared_logger.exception("exception raised", exc_info=e)
+            shared_logger.exception(
+                "exception raised",
+                exc_info=e,
+                extra={
+                    "summary": summarize_lambda_event(lambda_event),
+                },
+            )
 
             return f"exception raised: {e.__repr__()}"
 
     return wrapper
+
+
+def summarize_lambda_event(event: dict[str, Any], max_records: int = 10) -> dict[str, Any]:
+    """
+    Summarize the lambda event to include only the most relevant information.
+    """
+    summary = {}
+
+    try:
+        records = event.get("Records", [])
+
+        for record in records:
+            event_source = record.get("eventSource", "unknown")
+
+            if event_source == "aws:sqs":
+                aws_sqs_summary = summary.get(
+                    "aws:sqs",
+                    # if the key does not exist, we initialize the summary
+                    {
+                        "total_records": 0,
+                        "records": [],
+                    },
+                )
+
+                # The body contains the SQS message payload which is the S3
+                # notification event encoded as a JSON string.
+                event = json_parser(record["body"])
+
+                # So users know if we included only a
+                # subset of the records.
+                aws_sqs_summary["total_records"] += len(event["Records"])
+
+                for r in event["Records"]:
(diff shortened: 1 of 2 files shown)


handlers/aws/utils.py
```diff
@@ -125,13 +125,87 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+
+                    # We limit the number of records to `max_records` to
+                    # avoid large log payloads.
+                    if len(aws_sqs_summary["records"]) < max_records:
+                        aws_sqs_summary["records"].append(r.get("s3"))
+
+                summary["aws:sqs"] = aws_sqs_summary
+
+    except Exception as e:
+        shared_logger.exception("error summarizing lambda event", exc_info=e)
```
emilioalvap, line 204: Should this error path be reflected on the returned value?

zmoog: Oh, you mean including the error happened during summarization in the log?

constanca-m: I think the fact there is so much happening, it is not very clear where the exception is suppose to be coming from. So if is only from the records, to just do:

```python
try:
    records = event.get("Records", [])
except Exception as e:
    shared_logger.error("Failed to retrieve 'Records'", exc_info=e)
    return summary
```

handlers/aws/utils.py
```diff
@@ -125,13 +125,87 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+                    # if the key does not exist, we initialize the summary
+                    {
+                        "total_records": 0,
+                        "records": [],
+                    },
+                )
+
+                # The body contains the SQS message payload which is the S3
+                # notification event encoded as a JSON string.
+                event = json_parser(record["body"])
```
constanca-m, line 165: `event` variable already exists in the function parameters. I don't see it using elsewhere apart from being used to get the `records` so this wouldn't cause any issues, but are you overwriting the variable intentionally?

handlers/aws/utils.py
```diff
@@ -125,13 +125,87 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+
+                # So users know if we included only a
+                # subset of the records.
+                aws_sqs_summary["total_records"] += len(event["Records"])
```
constanca-m, line 169: I don't understand this very well. When would this happen? So if the summary does not exist, then you initialized it at 0, but if it exists, wouldn't this variable already be correct?

And if not, shouldn't it be changed inside the `for` cycle right below for when you append a new record?

constanca-m: Ok, I think I have just realized exactly what you are doing. So `total_records` is the total records there, but `records` is only up to the maximum (so 20).

handlers/aws/utils.py
zmoog: Ouch, I did a bad thing here. Fixing it!

zmoog: `notifications` is probably a better name:

```python
                # The body contains the SQS message payload which is the S3
                # notification events encoded as a JSON string.
                notifications = json_parser(record["body"])


```

handlers/aws/utils.py
zmoog: Yeah, I don't want to dump all the records from the lambda event into the log, I don't know how many of them we could receive. 

The lambda can receive up to 10 SQS messages and each message can contain multiple notifications. I want to protect and cap the log message size.

In the resulting log event:

```yaml
    "summary": {
        "aws:sqs": {
            "records": [ ... ],
            "total_records": 1
        }
```

- `records` contains the actual records
- `total_records` contains the total number of records in the lambda event

I don't feel this as confusing, but something like `first_{max_records}_records` is much better (I am using `first` because the summary collects the first max_records records in the lambda event).

handlers/aws/utils.py
zmoog: I can come from at least two sources:

- `json_parser()` fails to parse the notification
- the notification does not contain the `records` key

This summarization feature is not business-critical, so it should fail quickly and not interfere with the function execution too much. We try to summarize the function event, but if the summarization fails at any point, we stop, log the error, and return quickly.

handlers/aws/utils.py
zmoog: If we only have 1 notification the end result is:

```json
  "summary": {
        "aws:sqs": {
            "first_20_records": [{"key": "AWSLogs/blah/key"}],
            "total_records": 1
        }
```

constanca-m: That looks perfect!

constanca-m (approved):
Thanks Maurizio, looks good!

handlers/aws/utils.py
zmoog: Here is a shot from real life:

image

handlers/aws/utils.py
```diff
@@ -125,13 +125,93 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+                    #           "key": "AWSLogs/1234567890/CloudTrail-Digest/"
+                    #         }
+                    #       }
+                    #     }
+                    #   ]
+                    # }
+
+                    # We stop adding records to the summary once we reach
+                    # the `max_records` limit.
+                    if len(aws_sqs_summary[first_records_key]) < max_records:
```
xrmx, line 200: Should you also break the loop when reaching that? If so you can reverse the check and add a break to avoid looping for nothing.

handlers/aws/utils.py
```diff
@@ -125,13 +125,93 @@ def wrapper(lambda_event: dict[str, Any], lambda_context: context_.Context) -> s
...
+                        aws_sqs_summary[first_records_key].append(r.get("s3"))
+
+                # Update the summary with the new information.
+                summary["aws:sqs"] = aws_sqs_summary
+
+    except Exception as e:
```
xrmx, line 206: Nit: Maybe exc instead of e

handlers/aws/utils.py
zmoog: I was continuing the loop to get the total number of records.

Here's the lambda event structure (with minor changes for readability):

```json
{
    "Records": [
        {
            "messageId": "059f36b4-87a3-44ab-83d2-661975830a7d",
            "receiptHandle": "AQEBwJnKyrHigUMZj6rYigCgxlaS3SLy0a...",
            "body": "{\"Records\":[{\"awsRegion\":\"eu-west-1\",\"eventTime\":\"2024-01-31T19:06:30.680Z\",\"eventName\":\"ObjectCreated:Put\",\"eventSource\":\"aws:s3\",\"eventVersion\":\"2.1\",\"s3\":{\"bucket\":{\"arn\":\"arn:aws:s3:::mbranca-esf-data\",\"name\":\"mbranca-esf-data\"},\"object\":{\"key\":\"AWSLogs/123456/I-do-not-exist\"}}}]}",
            "attributes": {
                "ApproximateReceiveCount": "1",
                "SentTimestamp": "1545082649183",
                "SenderId": "AIDAIENQZJOLO23YVJ4VO",
                "ApproximateFirstReceiveTimestamp": "1545082649185"
            },
            "messageAttributes": {
                "myAttribute": {
                    "stringValue": "myValue", 
                    "stringListValues": [], 
                    "binaryListValues": []
                }
            },
            "md5OfBody": "e4e68fb7bd0e697a0ae8f1bb342846b3",
            "eventSource": "aws:sqs",
            "eventSourceARN": "arn:aws:sqs:us-east-2:123456789012:my-queue",
            "awsRegion": "us-east-2"
        }
    ]
}
```

So my goal is to provide the total number of the following elements (hoppa `jq` style):

```text
.Records[].body.Records
```

zmoog: oh, I see you point now, we can skip the inner loop 🤦

handlers/aws/utils.py
zmoog: @emilioalvap, did I address your suggestion?

PR #866, opened 2025-02-25
[docs] Add the new docs CI checks 

colleenmcginnis:
Adds the new docs CI checks. This will enable a preview to build our new Markdown-based docs.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,0 +1,19 @@
+name: docs-build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request_target: ~
+  merge_group: ~
+
+jobs:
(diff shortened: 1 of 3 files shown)


kuisathaverat:
I have fixed the lint here to avoid having to make changes in a new PR. The litter used in this project is not static and the rules can change between executions. The lintter must be static to avoid this kind of issue. The lintter is build on each execution takes about 1 minute to build and a few seconds to make the lint. It lints all the files in the repo, not only the changes in the PR. I suggest moving to a proper tool to make the lines like [pre-commit](https://pre-commit.com/index.html) that supports most of the tolling used in that container but in a much better way.

alaudazzi (approved):
LGTM

